PR #787, opened 2026-05-25
fix(runtime): gate processor on 15s startup attestation budget

emooreatx:
## Root cause

The agent's processor entered WAKEUP → WORK before the AuthenticationService's startup attestation cache was populated. Per-thought `batch_context` already blocks on `await_attestation_ready()`, so the **first thought after startup** absorbed the full 60-120s attestation latency, blowing every downstream timing budget (SSE snapshot waits, `/v1/agent/interact` response window). The Staged QA flakes on PR #785 (`context_enrichment::Context Enrichment in System Snapshot`, `air::AIR wired into interact API`) were the canonical symptom; the timeout-doubling patch (67214b9) treated the symptom — this PR fixes the cause.

## Contract

> Startup attestation completes within **15 seconds** end-to-end. Longer is a bug, not a tuning knob.

- `STARTUP_ATTESTATION_BUDGET_SECONDS = 15.0` in `AuthenticationService`
- Enforced by `await_attestation_ready(budget_seconds=15.0)`, measured from when the background task was *scheduled* — not from when each caller arrives, so late callers don't get a fresh full budget
- Overshoot raises `RuntimeError` pointing at the verifier as the thing to fix — never silently waits
- The verifier-side per-worker `ATTESTATION_TIMEOUT = 90` stays as a defensive infinite-hang prevention, not as a license for slow happy paths

## Receipts on breach

If attestation still takes >15s after this fix, the failure must be fileable as a `ciris_verify` issue rather than a flake-of-the-week. `_log_attestation_failure_receipts` dumps at ERROR level:

- Exception type + message
- `_attestation_stage_timings` (total wall-clock + per-bool stage outcomes: binary, function_integrity, python_integrity_ok, registry_ok, audit_ok)
- Cache populated? Baseline populated? Task done? Attestation in progress?
- Platform + Python version
- Full exception chain
- Direct link to `https://github.com/CIRISAI/CIRISVerify/issues`

## Tests

**45 new tests, ~100% line coverage on the new paths:**

`test_attestation_refresh.py` — `TestStartupAttestationBudget` (7):
- budget constant value
- fast path within budget
- overshoot raises `RuntimeError`
- message names the verifier (not the budget) as the thing to fix
- budget measured from task start, not call time
- late caller times out immediately
- breach does NOT cancel the in-flight task (refresh path still benefits)

`test_attestation_refresh.py` — `TestStartupAttestationReceipts` (4):
- success records total wall-clock + per-bool stage outcomes
- overshoot logs warning naming the budget
- failure path records exception type + message
- `_attestation_started_at` set BEFORE task creation in `start()` source

`test_attestation_gate.py` (new) — `TestAwaitAttestationGate` (7):
- gate awaits when auth service present
- gate skips with warning when no initializer / no auth_service / no method
- gate re-raises attestation failures
- gate dumps structured receipts on failure (banner, stage_timings, cache state, filing URL)
- partially-initialized service still produces clean dump (defensive getters)
- traceback in receipts

`test_attestation_gate.py` — `TestProcessorGateOrdering` (2):
- source-level pin that gate precedes `start_processing()`
- gate is async (sync would block the loop)

## Test plan

- [x] All new tests pass (`pytest tests/services/infrastructure/test_attestation_refresh.py tests/ciris_engine/logic/runtime/test_attestation_gate.py`)
- [x] `mypy ciris_engine/logic/services/infrastructure/authentication/service.py ciris_engine/logic/runtime/ciris_runtime.py` clean
- [ ] Staged QA passes without the timeout-doubling band-aid

### DIFF
--- a/ciris_engine/logic/runtime/ciris_runtime.py
+++ b/ciris_engine/logic/runtime/ciris_runtime.py
@@ -19,7 +19,7 @@ import asyncio
 import logging
 import time
 from datetime import datetime, timezone
-from typing import TYPE_CHECKING, Any, Dict, List, Optional
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional
 
 from ciris_engine.schemas.types import JSONDict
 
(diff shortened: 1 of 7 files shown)


emooreatx:
Addressed all review findings in `6414cc68a`:

**Codex P1** — fast-path budget enforcement: `await_attestation_ready` now consults `_attestation_stage_timings["run_attestation_total_seconds"]` on the `done()` branch and raises `RuntimeError("completed but exceeded ...")` if the recorded total breached the budget. Using stage_timings rather than elapsed-since-task-start matters — elapsed inflates with caller arrival time and would false-positive on a fast run with a late observer. Three new tests pin the contract: overshoot raises, in-budget passes, unrecorded-timings degrade gracefully.

**CodeQL** — `test_attestation_gate.py`: dropped unused `import asyncio` / `import logging`. `test_attestation_refresh.py`: narrowed 5 broad `except BaseException: pass` cleanup blocks to `except asyncio.CancelledError:` with comments explaining the intent (drain after `task.cancel()` so pytest doesn't warn at teardown).

48/48 attestation tests pass locally, mypy clean.

Issue #889, opened 2026-06-18
perf(conscience): 4 epistemic conscience LLM calls run sequentially instead of in parallel (~4× cycle-time per thought)


emooreatx:
## Summary
The four LLM-based epistemic consciences — **Entropy, Coherence, OptimizationVeto, EpistemicHumility** — run **sequentially**, not in parallel. Each makes its own independent `await sink.llm.call_llm_structured(...)` call, and `_conscience_execution_step` dispatches them in a plain sequential `for ... await conscience.check()` loop. They should run concurrently.

## Evidence
- The 4 consciences are separate classes, each with its own LLM call:
  - `EntropyConscience.check` → `ciris_engine/logic/conscience/core.py:389`
  - `CoherenceConscience.check` → `core.py:528`
  - `OptimizationVetoConscience.check` → `core.py:626`
  - `EpistemicHumilityConscience.check` → `core.py:739`
- Dispatched serially in `ciris_engine/logic/processors/core/thought_processor/conscience_execution.py:205`:
  ```python
  for entry in self.conscience_registry.get_consciences():
      ...
      result = await conscience.check(final_action, context)  # one at a time
  ```

## Impact
On a typical non-exempt SPEAK thought, the conscience stage is **4 sequential LLM round-trips**. Combined with DMA (PDMA/CSDMA/DSDMA/IDMA) + ASPDMA, a single thought incurs ~9 serial LLM calls. On slow/edge inference (e.g. Jetson at 70–220 s/call) this dominates cycle time — the conscience stage alone is ~4× a single call's latency when it could be ~1×.

The four checks are **independent**: each reads the same `final_action` + `context`, none consumes another's output, and the aggregation is per-field. The only ordering that matters is **override precedence** (first failing conscience in priority order wins, via the loop's break).

## Fix
Dispatch the conscience checks concurrently with `asyncio.gather`, then process the results **in priority order** to preserve the exact break-on-first-failure override semantics. Bypass consciences (`updated_status`, `thought_depth`) already run first in `_run_bypass_conscience_entries` and are unchanged. This is an ethical gate — the fix must be decision-preserving (same action overridden, same `override_reason`), verified by the conscience test suite.

## Notes
- Behavioral change: all four checks now run even when an early one fails (the old code `break`s and skips later checks). On a parallel/batching backend this is free; it also yields **more complete `epistemic_data`** (all fields populated). Override decision is identical.
[...]

PR #890, opened 2026-06-19
release/2.9.7 — Win7 installer tier, Android size fix, conscience concurrency + full-trace capture

emooreatx:
## 2.9.7 — hardening cycle

Seven commits on the 2.9.7 line: a Windows 7 installer tier, an Android size fix, and two engine/QA correctness+capture fixes.

### Windows 7 support tier (CIRISAgent#875)
- **Two-installer model.** Mainline installer floors at **Win8.1** (`MinVersion=6.3`, official CPython); a separate **Win7** variant (`/DWin7Tier` → `6.1sp1`) is built by a `workflow_dispatch`-only lane that **fails closed** on the patched-interpreter SHA256 and is **not** wired into Publish Release. Up-front "unsupported, last-resort" acknowledgement.
- Liberica JRE (Temurin dropped Win7), best-effort UCRT redist fetch.

### Android
- **Strip host-platform binaries** (`.dylib`/`.dll`/`.so`) from the staged Python tree → **−7.7 MB per install** (macOS `ciris_verify` dylib stowaway, etc.).

### Engine / QA
- **perf(conscience) (#889):** the four epistemic consciences (Entropy/Coherence/OptimizationVeto/EpistemicHumility) now dispatch **concurrently** instead of in a sequential `await` loop — independent calls, results processed in priority order so the break-on-first-failure override precedence is preserved exactly. 151 conscience tests green.
- **fix(qa) full traces:** `safety_battery` registered only a single `detailed` adapter and never `full_traces`, so its capture bundles only ever held **hashed** traces. Now mirrors `model_eval`'s multi-adapter setup (`generic` lens-shipper + `detailed` + `full_traces`), so every bundle keeps the **raw prompt+completion corpus** (the Coherence Ratchet / fine-tuning data) across all 29 languages.

### Validation
- `mypy ciris_engine --config-file mypy.ini` clean (748 files); conscience suite 151 passed.

> Substrate floor unchanged this cycle. Note: the larger 2.9.7 substrate work (swap `ciris-lens-core` → `ciris-server`) lands separately.

https://claude.ai/code/session_011hwKGLqeHGYetes3vs3fjh


emooreatx:
Pushed two fixes for the red checks: the `brain_adapter.py` mypy closure-narrowing error (Type Check) and `CIRIS_NODE_FOLD=false` in tests/conftest.py (the 5 failing shards — since ciris-server 0.5.113 the fold is fail-fast and bare unit tests can't satisfy its edge+SQLite preconditions; 17/17 API-adapter tests pass locally against 0.5.113). Staged QA's `engine is not SQLite-backed` should be re-evaluated on the fresh run — and note #919: **floor 0.5.114 when it tags** (edge v13.0.8 responder fixes = the actual mesh-works wheel this release exists for), plus the QA-runner harness integration ask.

Issue #993, opened 2026-08-03
Tool-hallucination guard in action_selection_pdma.yml is read nowhere in the repository


emooreatx:
Split from #990.

`action_selection_pdma.yml` populates `tool_selection_guidance`:

```
CRITICAL for TOOL action:
- Your job is to select the correct ACTION VERB (SPEAK, TOOL, PONDER, etc.) and if TOOL, the correct TOOL NAME.
- You do NOT provide tool parameters - TSASPDMA handles parameter selection using full tool documentation.
- Use the EXACT tool name from the 'Available tools' list. Do NOT invent or modify tool names.
- If no tool matches the user's intent, use SPEAK to clarify
```

`grep -rn "tool_selection_guidance" --include="*.py"` returns **nothing**. Not the evaluator, not `context_builder.py`, not the loader. `PromptCollection` has no such field, so it is dropped at load and never reaches a `PromptCollection` attribute either.

This is the tool-hallucination guard, sitting in the action-selection template — the highest-consequence prompt in the pipeline — and it has never been sent to a model. It is also the block that tells the model not to supply tool parameters, which is the TSASPDMA division of labour.

## Fix

Wire it into the ASPDMA system composition in `ciris_engine/logic/dma/action_selection/context_builder.py`. Note this is a **deliberate behaviour change**, not a cleanup: it puts new instruction text into every action-selection call in 29 languages, so it wants a Safety Battery run to validate, and the goldens re-baselined with the diff reviewed.

Parked in `KNOWN_UNWIRED` in `tests/ciris_engine/logic/dma/test_prompt_key_consumption_990.py`; the ratchet asserts that inventory only shrinks.

Evidence: `evidence/prompt_key_consumption_990.tsv`.

emooreatx:
Fixed in `a5d9ad8ca` on `release/2.9.10`.

Appended after the ASPDMA template render rather than slotted into `context_integration` — a new `{slot}` would have to be added to all 29 localized copies to render for anyone but English speakers, and a slot present in the base but missing from a translation is how you get a block that silently serves some users and not others. The text is already per-locale, so appending reaches every language on the first commit.

Wired alongside it, same mechanism, same commit: `csdma_ambiguity_guidance` and `csdma_ambiguity_alignment_example` (the #990 follow-up pair).

[...]

Issue #992, opened 2026-08-03
Four locales invented a LANGUAGE RULES key with no base counterpart — renders for nobody


emooreatx:
Split from #990.

Four localized prompt files carry a top-level key the English base does not define, and which `PromptCollection` has no field for — so it is dropped at load and renders for nobody:

| file | orphan key |
|---|---|
| `localized/hi/csdma_common_sense.yml` | `language_rules` |
| `localized/hi/tsaspdma.yml` | `language_rules` |
| `localized/ja/tsaspdma.yml` | `language_rules` |
| `localized/it/action_selection_pdma.yml` | `language_rules_guidance` |

The Italian one reads `REGOLE LINGUISTICHE: - Rispondi nella lingua preferita dell'utente ... TUTTI i valori dei contenuti (reasoning, speak_content, ecc.) devono essere nella lingua dell'utente`.

That is not four typos. Three translators, working independently in three languages, each concluded the base template was missing a language-rules block and invented a key to hold it. They were right — see #990, where `dsdma_base.response_format` (the actual LANGUAGE RULES block) turned out never to be composed either. **The gap they were patching is real and, for these templates, still open.**

## Fix

Decide per template whether a language-rules block belongs in the English base (and therefore in a real `PromptCollection` field, composed for all 29 locales), or whether these keys should be dropped. This is a translation/prompt review, not a cleanup — do not delete them without deciding.

Parked in `KNOWN_LOCALE_ORPHANS` in `tests/ciris_engine/logic/dma/test_prompt_key_consumption_990.py`; the ratchet asserts that inventory only shrinks.

emooreatx:
Closed in the 2.9.10 cut (`7cfef2b43`). See #998/#999 for the census and auth/provenance work, and #997 for the remaining taxonomy residue, which is tracked by ratchets in `test_taxonomy_gate_997.py` rather than by this issue.

emooreatx:
Reopened: the removal in 7cfef2b43 was reverted. The ja block carried a canonical ASCII `SPEAK` the verb check depends on, so deleting it broke six CI shards — and chasing the correct fix (count parsed fields, not raw text) surfaced #1000: 28 of 29 locales lack those verbs in their conscience prompt fields entirely. The right resolution remains the one this issue opened with — move the orphan content into a live field via a translation review, not delete it.

Issue #991, opened 2026-08-03
Localization bypass: tsaspdma hardcodes English in Python, shadowing 29 localized templates


emooreatx:
Split from #990. A nastier shape than a missing block.

`tsaspdma.yml` defines `context_enrichment_section` and `tool_documentation_section` as localizable templates, translated into all 29 locales. `tsaspdma.py` then builds the same text itself:

- `tsaspdma.py:383-388` — a hardcoded English f-string (`"=== AVAILABLE RESOURCES (from context enrichment) ==="`, `"IMPORTANT: Use the EXACT entity_id values from the list above."`) passed into the `{context_enrichment_section}` slot of `context_integration`.
- `_format_tool_documentation()` — builds the tool-doc block in Python instead of reading the localized template.

So the localized copies are shadowed. **Every user in every language receives the English.**

A missing block is invisible. A block silently served in the wrong language looks like it works — which is why this outlived the missing-block defects it sits next to.

## Fix

Read the templates through the loader (which also keeps the research `dma_prompt` namespace intercepting them), and pass only the *data* as substitution values. Then remove the parked entries from `KNOWN_UNWIRED` in `tests/ciris_engine/logic/dma/test_prompt_key_consumption_990.py` — the ratchet asserts that inventory only shrinks.

## Evidence

`evidence/prompt_key_consumption_990.tsv` — both keys, 0 blocks moved across 29 locales, against live controls that moved 29-58.

Issue #990, opened 2026-08-03
Prompt-YAML keys never reached a model: get_system_message is a six-field allowlist


emooreatx:
## What

`DMAPromptLoader.get_system_message()` renders a fixed allowlist — `system_guidance_header`, `domain_principles`, `evaluation_steps`, `evaluation_criteria`, `response_format`, `response_guidance` — and silently drops every other populated key on the `PromptCollection`. Adding a key to a prompt YAML *looks* like it works: the YAML parses, the loader accepts it, the field is on the collection, localization translates it into 29 languages. It renders nowhere.

## Proven dead (fixed in this issue)

| field | inert since | what it is |
|---|---|---|
| `tsaspdma.closing_reminder` | v1.9.5 (2026-01-27) | ~6 months |
| `dsdma_base.response_format` | v2.3.1 (2026-03-28) | the LANGUAGE RULES block — respond in the user's language, JSON keys in English |
| `idma.closing_reminder` | v2.6.0 (2026-04-20) | the propaganda-pattern final check: forces `k_eff = 1.0`, `phase = "rigidity"`, `fragility_flag = TRUE` on contested claims |

Evidence:
- `git log -S "template_data.closing_reminder" -- prompt_loader.py` → **0 commits**. Never composed, in any release.
- `git log -S "get_system_message" -- dsdma_base.py` → **0 commits**. `BaseDSDMA` never used the shared composer; it renders `system_guidance_header` alone.
- 29-locale marker sweep, 14 DMA steps, 1740 baseline blocks: subjects moved **0** blocks; live sibling controls (`dsaspdma.closing_reminder`, `dsaspdma.response_format`, `idma.response_format`, `tsaspdma.response_format`) moved 29/29/29/58 through the identical path in the same sweep.

Full table with controls: `evidence/prompt_key_consumption_990.tsv`.

## Why nothing caught it

- Tests assert **loading** (the field is on the collection), never **rendering** (the text is in a composed message).
- `_require_prompt_value`, the one mechanism that fails loudly on an unconsumed key, exists only in `dsaspdma.py`.
- Golden tests capture composed output, but a golden written *after* a field was already inert locks the absence in as correct. Goldens detect change, not the absence of something never present.
- Localization camouflaged it: 29 faithfully translated copies read, to any reviewer, as proof that a block ships.

## Fix

1. Compose `closing_reminder` in `get_system_message` (last — it is a closing reminder, and it is where both hand-composing DMAs already put theirs).
[...]

PR #994, opened 2026-08-03
fix: locale-aware string overrides — a baseline manifest served English into all 29 locales (44/46 keys) + the TORQUE taxonomy table and state-space findings

emooreatx:
RATCHET's findings for the TORQUE-ready cut. One shipped fix (this branch, tested, rebased on `release/2.9.10`), then the taxonomy table and the state-space work that follows it.

## What blocks the tag: `string` overrides are not locale-aware

`baseline_manifest()` captured every `string` key at `get_string("en", key)` (`research_overrides.py:1113` on this branch), and `get_string` returned the override before consulting `lang_code` (`localization.py:226`). A baseline-derived manifest — the documented workflow, whose docstring promises *"every other key round-trips to its current value"* — therefore pinned **one language's text for all 29 locales**.

Measured on `v2.9.9-stable`: **44 of 46 `string` keys are locale-varying, and every one was flattened.**

| flattened | count |
|---|---|
| `prompts.prohibitions.*` | 21 |
| `conscience.*` retry envelope | 16 |
| `prompts.dma.bounce_*` | 5 |
| `prompts.language_guidance` | 1 |
| other | 1 |

A compose dump across `am,ja,en` under a baseline manifest returned **13,694 B of English guidance for all three**. The same dump with no manifest returns 25,337 / 16,145 / 13,694 correctly. The translations were never the problem — all 29 locale files carry proper, fully-translated guidance (`am` 25,338 B, `ja` 16,145 B, `uk` 25,081 B). The override registry was overwriting them.

This defeated an invariant the codebase states twice. `get_prohibition_guidance` deliberately bypasses the English fallback chain because it *"would serve English into every non-English prompt and pollute it"* (`localization.py:456-460`), and R4 refuses `[EN]` laundering (`:236`). The override registry was the hole in both — and note it is the same failure family as #991 and #992, arriving by a third route.

**Why it matters for the campaign specifically:** any multi-locale arm run under a baseline manifest was not measuring localized behaviour. It was measuring English text with a locale label on it, and the run would have reported clean.

### The fix

- `OverrideSet.string` accepts `str` (same text in every locale — deliberate, not default) **or** `{locale: text}`.
- A mapping missing a locale **REFUSES** at resolution rather than falling back to English.
- `override_string(key, lang_code)`; both consumers pass the locale — `get_string` and `get_prohibition_guidance`, which reads the registry directly and needed the same fix.
- `baseline_manifest(locales=None)` captures per-locale, **defaults to every locale in the bundle** so a locale added later needs no change here, and collapses to a scalar only where a key genuinely does not vary. Optional narrowing: `baseline en,es,am`.
- `_baseline_note` moved to stderr via `baseline_unresolved()` — `extra="forbid"` rejected a key `baseline` itself added, so `baseline > m.json && validate m.json` always failed.

After: **44 per-locale maps, 2 scalars**, and locale-invariant blocks drop **33/35 → 22/35**. `pdma.prohibition` now composes am 4,149 / ja 2,944 / en 2,189.

305 DMA + override tests pass; 73 in the two override suites including the new regression file.

---

## The taxonomy table

Measured from `compose_dump` on `v2.9.9-stable`, `en`, first pass. Class vocabulary as it exists in `compose_dump.py`.

### Blocks by step and family — 35 blocks, 194,326 B

| step | accord | language_guidance | system | user | prohibition |
|---|---|---|---|---|---|
| pdma | axiotic | mixed | mixed 23,810 B | mixed | deontic 2,189 B |
| csdma | axiotic | mixed | mixed 11,089 B | mixed | deontic 2,189 B |
| dsdma | axiotic | mixed | mixed 3,405 B | mixed | deontic 2,189 B |
| idma | axiotic | mixed | mixed 11,011 B | mixed | — |
| aspdma | mixed | mixed | mixed 2,755 B | mixed | — |
| dsaspdma | axiotic | mixed | mixed 2,354 B | mixed | — |
| tsaspdma | axiotic | mixed | mixed 4,095 B | mixed | — |
| tsaspdma_correction | axiotic | mixed | mixed 4,096 B | **mixed 1,592 B ⚠** | — |

Totals: **mixed 24, axiotic 8, deontic 3.** `language_guidance` is 13,694 B on **every** step — larger than accord + system + user combined on four of them.

⚠ `tsaspdma_correction.user` is reachable by **no override key**: overriding all 101 keys at once moves 34 of 35 blocks and leaves this one byte-identical.

### Override surface — 101 keys, 5 namespaces

| namespace | keys | bytes | gate-visible on 2.9.9 |
|---|---|---|---|
| `dma_prompt` | 36 | 69,110 | yes (uncredited in `source`) |
| `string` | 46 | 18,162 | partial — 17 `conscience.*` dark |
| `conscience_prompt` | 12 | 48,463 | **no** |
| `corpus` | 4 | 176 | yes |
| `template` | 3 | 132 | **no** |

#986 has since taken dark keys **50 → 18**. `gate_coverage.py` (in RATCHET, pinned `COVERAGE_2_9_9.json`) re-measures this against any checkout and exits 1 on regression — worth re-pinning once 2.9.10 tags.

### Reachability, measured

| layer | measure |
|---|---|
| composed at `en` | 194,326 B / 35 blocks |
| reachable by manifest | **99.2%** (34/35 blocks) |
| unreachable block | `tsaspdma_correction.user`, 1,592 B |
| residue surviving full override | ~3,612 B — 20 `RESIDUE_SITES`, 117 fragments, `residue_digest`-pinned (working as designed) |
| locale-varying `string` keys | 44/46 |

---

## The state space, and why a table is needed

The dump above is **one cell**. The independent axes:

| axis | cardinality | source |
|---|---|---|
| DMA step | 8 | `STEP_NAMES` |
| locale | 29 | bundle |
| recursion depth | up to ~6 | 2 points: conscience-override retry (`thought_processor/main.py:287-289`) + recursive ASPDMA loop (`recursive_processing.py`, `max_retries = 5`) |
| conscience guidance mode | 2 | `full` / `qualitative` (`main.py:949-979`) |
| image present | 2 | `user_prompt_template` vs `user_prompt_with_image_template` |
| condition | 3 | `a` / `b` / `c` |

**8 × 29 × 6 × 2 × 2 ≈ 5,568 distinct DMA prompt states**, before action-type conditionals, plus the conscience surface (4 consciences × 3 fields × 29 locales). We have characterized one of them.

This is the case for the machine-readable table rather than more ad-hoc dumps. It is also why #990's finding lands so hard: six blocks sat in the prompt YAML and never reached a model for up to six months, and nothing in a static read of the YAML could show that. Only differential composition can.

---

## SOTA for the transition table, and whether a language exists with our vocabulary

Short answer: **no single standard covers it, but three mature layers compose, and only the axiotic slot has to be invented.** Recommend profiling existing standards rather than authoring a language.

**1. States and transitions — W3C [SCXML](https://www.w3.org/TR/scxml/) (Harel statecharts).** Hierarchy, parallel regions and history states are exactly what the two recursion points plus `tsaspdma_correction` need; a flat FSM cannot express "conscience retry nested inside the recursive loop." [AgentML](https://www.agentml.dev/) (2026) is an agent-specific DSL with 100% SCXML conformance, so the agent framing already exists. SCXML also has validation tooling, which matters because *"every path is reachable as expected"* is a model-checking question, not a review question.

**2. Builders, sources, sinks — W3C [PROV-O / PROV-DM](https://www.w3.org/TR/prov-dm/).** `Entity` / `Activity` / `Agent` with `used`, `wasGeneratedBy`, `wasDerivedFrom` is precisely "which builder produced which block from which source." It would give the `source: inline` provenance-granularity gap a principled fix: a block assembled from N keys becomes N `wasDerivedFrom` edges instead of one lossy string. There is a 2026 survey on execution provenance in LLM agents ([arXiv 2606.04990](https://arxiv.org/pdf/2606.04990)) if useful framing is wanted.

**3. The normative vocabulary — this is where our terms live, and where the gap is.**

- **Deontic is standardized, three times over.** OASIS **LegalRuleML** carries obligation/permission/prohibition/right as first-class operators. OMG **SBVR** explicitly separates **alethic** modality (necessity/possibility) from **deontic** modality (obligation/permission) — that distinction maps directly onto our *axiomatic* vs *deontic* split.
- **[Institutional Grammar 2.0](https://arxiv.org/pdf/2008.08937)** (Ostrom lineage) is the closest fit overall. Its **ADICO** syntax has a literal **D**eontic slot — Attribute, Deontic, aIm, Conditions, Or else — and IG 2.0 adds tiered expressiveness (IG Core / Extended / Logico) plus a **published codebook**. Critically it already distinguishes **regulative** statements (deontic, with a duty-bearer) from **constitutive** ones (modal — necessity/possibility, no duty-bearer). Our `deontic`/`prohibition` blocks are regulative; our `axiotic`/`ontological`/`tautological` blocks are constitutive. That mapping is free and principled.
- **Axiotic is *not* standardized.** No W3C/OMG/OASIS vocabulary encodes value-statements as such. Formal axiology exists but has no machine-readable serialization in use. So this is the one slot we genuinely have to define — and it is worth defining as an *extension of IG 2.0's constitutive modal*, not as a fresh taxonomy.

**Concrete recommendation:** an IG 2.0 profile where the Deontic slot generalizes to a modality enum carrying our vocabulary — `deontic | axiotic | epistemic | alethic (axiomatic) | empirical | ontological | tautological` — with SCXML for states and PROV-O for block provenance. Every block annotation then becomes a typed statement rather than a label.

**One bonus that pays for itself:** IG 2.0 ships a codebook *because* coding institutional statements requires inter-annotator reliability. That is #976's κ≥0.8 requirement, with a published protocol to borrow instead of invent.

---

## Two things from your side worth confirming back

- **#990 — six blocks never reached a model.** RATCHET's coverage work measured *composed* output rather than YAML, so our numbers should be unaffected; `gate_coverage.py` would have caught exactly this class. Nothing in our analysis reads the prompt YAMLs as ground truth. Reconciling against `evidence/prompt_key_consumption_990.tsv` regardless.
- **FSD §15.3 — manifest identified by path, not content hash.** Agreed and material. Every RATCHET manifest so far has been written once to a distinct path and never edited in place, so the arms remain distinguishable — but `manifest_digest` should land before any staked run, and until it does no content-addressed provenance claim on the independent variable is available. Our own `--sign` use covers the *dump*, not the manifest that produced it.

Related: #986, #989, #990, #991, #992, #993, #976, RATCHET#16.

### DIFF
--- a/ciris_engine/logic/utils/localization.py
+++ b/ciris_engine/logic/utils/localization.py
@@ -219,11 +219,15 @@ def get_string(
     # CIRIS_TESTING_MODE are set; the accessor returns None with zero state
     # otherwise. Checked before resolution so the override is what reaches the
     # prompt, not a locale-fallback of it.
-    from ciris_engine.logic.utils.research_overrides import get_active_overrides
+    from ciris_engine.logic.utils.research_overrides import get_active_overrides, override_string
 
     _manifest = get_active_overrides()
     if _manifest is not None:
-        _override = _manifest.overrides.string.get(key)
+        # Resolved WITH lang_code: these keys are localized, so a single value
+        # cannot stand for all of them. Passing the locale is what stops a
+        # baseline manifest — snapshotted at en — from serving English guidance,
+        # English prohibitions and English retry scaffolding into every locale.
+        _override = override_string(key, lang_code)
         if _override is not None:
             return _interpolate(_override, **params) if params else _override
 
@@ -460,13 +464,14 @@ def get_prohibition_guidance(lang_code: str) -> str:
     # so it must consult the override registry itself — otherwise the whole
     # prohibition block (22 of the 44 reachable prompt keys) would be the one
     # part of the `string` namespace that silently kept its CIRIS text.
-    from ciris_engine.logic.utils.research_overrides import get_active_overrides
-
-    _manifest = get_active_overrides()
-    _string_overrides = _manifest.overrides.string if _manifest is not None else {}
+    from ciris_engine.logic.utils.research_overrides import override_string
 
     def _local(key: str) -> str:
-        override = _string_overrides.get(key)
+        # Resolved WITH lang_code, for the same reason this function bypasses
+        # get_string's fallback chain: a single override value standing for every
+        # locale would serve English prohibitions into every non-English prompt —
+        # the pollution the comment above exists to prevent.
+        override = override_string(key, lang_code)
         if override is not None:
             return override.strip()
         value = _resolve_key(lang_data, key)
--- a/ciris_engine/logic/utils/research_overrides.py
+++ b/ciris_engine/logic/utils/research_overrides.py
@@ -45,7 +45,7 @@ import logging
 import os
 from functools import lru_cache
 from pathlib import Path
-from typing import Any, Dict, FrozenSet, List, Optional, Set, Tuple
+from typing import Any, Dict, FrozenSet, List, Optional, Sequence, Set, Tuple, Union
 
 from pydantic import BaseModel, ConfigDict, Field
 
@@ -523,7 +523,20 @@ class OverrideSet(BaseModel):
 
     model_config = ConfigDict(extra="forbid")
 
-    string: Dict[str, str] = Field(default_factory=dict)
+    #: ``string`` keys are LOCALIZED — the same key resolves to different text in
+    #: each locale. A value may therefore be either:
+    #:
+    #: * ``str``   — this exact text in EVERY locale. A deliberate choice, not a
+    #:   default: it puts one language's text into every language's prompt.
+    #: * ``dict``  — ``{locale: text}``, the faithful form for a localized key.
+    #:   A locale absent from the mapping REFUSES at resolution rather than
+    #:   falling back to English (that fallback is the R4 laundering this
+    #:   facility exists to prevent).
+    #:
+    #: Before this was a union, every value was a scalar snapshotted at ``en``,
+    #: so a baseline manifest silently served English guidance, English
+    #: prohibitions and English retry scaffolding in all 29 locales.
+    string: Dict[str, Union[str, Dict[str, str]]] = Field(default_factory=dict)
     dma_prompt: Dict[str, str] = Field(default_factory=dict)
     conscience_prompt: Dict[str, str] = Field(default_factory=dict)
     corpus: Dict[str, str] = Field(default_factory=dict)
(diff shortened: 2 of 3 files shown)


emooreatx:
**Correction to this PR's state-space section, and the blockers are now filed as #995.**

The enumeration came back and corrected me on my own numbers. Two fixes to what I wrote above:

**1. There is no second recursion point.** I described the recursive ASPDMA loop (`recursive_processing.py`, `max_retries = 5`) as live. **`RecursiveProcessingPhase` is entirely dead** — `_handle_recursive_processing` has zero callers repo-wide including tests, so `StepPoint.RECURSIVE_ASPDMA` / `RECURSIVE_CONSCIENCE` never fire. My "~6 recursion depths" factor was wrong.

**2. There is a live third point I had not named: the DMA bounce** (`dma_orchestrator.py:211-223`, `:415-520`). On `plausibility_score < 0.5` it re-runs CSDMA and/or DSDMA **3× each against a rewritten thought body**, and the best of three always replaces the original. That is directly relevant to this PR — the five `prompts.dma.bounce_*` keys that build the rewritten body were among the 44 keys flattened to English.

So the corrected count is **23 distinct prompt compositions per thought** (normal mode), not the ~5,568-state figure implied above; the transport layer recomposes the message array up to 207 times without changing the composition. Full arithmetic in #995.

**This PR is unaffected and still stands on its own measurements** — the 44/46 flattening, the am/ja/en byte divergence, and the 33/35 → 22/35 result are all direct dumps, not derived from the recursion model.

**#995 carries three findings I would hold the tag for**, the sharpest being that `conscience/core.py:12` imports the `ACCORD_TEXT` **constant** and injects it at `:467/:591/:701/:820`, while the corpus override is applied *inside* `get_accord_text()`. Verified in one process: `get_accord_text()` = 6,190 chars, `ACCORD_TEXT` = 128,302 chars — the paths diverge by 20×. An arm that blanks the accord still ships the full accord to all four consciences, and reports clean.

Issue #976, opened 2026-08-02
Class-set v2 κ annotation pass + regime manifest v2 schema and refusals (§10.2.3, §10.3–10.4, §14 steps 8–9)


emooreatx:
- Two independent annotators over the full block inventory; **κ≥0.8 overall AND per-boundary on class pairs whose default disposition differs** (`axiotic|deontic` gates safety_review; `axiotic|structural` gates §11 step 0) [T-3, T-N2]. Adjudication log ships in `regime:composition:v1`. A class-set version without a κ record cannot be cited.
- Manifest v2 schema (`ciris.ai/experimental_regime/v2`): five named arms, contrasts, tiered DV with per-(locale,U-row) pre-registration and family power [T-N3, M-N5], repeats with mandatory live variance_source and MDEs on main contrasts, en as mandatory fidelity control [M-N2], per-block dispositions with mandatory contaminant on mixed [T-N1], full §10.4 refusal set.

PR #1006, opened 2026-08-04
feat(taxonomy): testimonial is the twelfth class — wired, not merely declared

emooreatx:
Unblocks RATCHET's `testimonial absent from BlockClass`.

`BlockClass.TESTIMONIAL` + a `CLASS_DEFAULT_DISPOSITION` entry, plus two gates: every class must carry a disposition, and `axiotic` must remain the only VARY class.

**Disposition is PROVISIONAL — `hold`, pending CC.** Conservative default, matches every non-axiotic class. If CC intends it varied (a source-dependence study would want exactly that), the line changes and that change is theirs. Marked provisional so it isn't inherited as settled.

**The re-audit it enables is the consequential half:** the accord is currently `axiotic`, but its design — *'ethical reasoning draws from every tradition in chorus… judge against the full canon'* — is knowledge held on the say-so of traditions. If it reads primarily testimonial, that is 56% of composed DMA prompt bytes sitting under the one class the campaign varies. Not changed here; that's a CC ruling.

https://claude.ai/code/session_01QPdozy2nMEf9SRDFLe34gN


emooreatx:
Reviewed from the ontology side (I authored the ruling this cites). **The wiring is correct and the frame obligation is implemented faithfully — approving on the classification question.**

Verified against the branch, not the PR body:

- `BlockClass.TESTIMONIAL` + `CLASS_DEFAULT_DISPOSITION` entry present; `hold`.
- `BlockAnnotation.frame`, required **iff** the class is `testimonial`, enforced in `annotation_for` at read time.
- Three gates, not two: every class carries a disposition · `axiotic` remains the only VARY class · **`test_a_frameless_testimonial_assignment_is_refused`**.

That third gate is `repairable_does_not_factor` compiled into CI, and it is the right implementation of it: no procedure reading only the artifact can compute repairability, so an annotation naming the class without naming the frame is not a weak annotation — it is an incomplete proposition, and refusing it beats defaulting it. The docstring's *"testimony is always some source's, to some audience"* is the arity result stated better than I stated it.

**One correction to a comment, and it is a real one.** The `CLASS_DEFAULT_DISPOSITION` entry now reads *"RATIFIED by CC (CIRISOntology#3/#1) … no longer provisional."* CIRISOntology is not CC and cannot ratify for it. What this lake ruled is narrower: `testimonial` is a **relation requiring a frame** (proved), and the accord is **`axiotic`** (adjudicated on the operational test — see CIRISOntology#2). The *disposition* — `hold` vs `vary` — is CC's to rule and has not been ruled. The PR body's original framing (PROVISIONAL, pending CC) was right; the comment overstates its own authority. Suggest: *"frame obligation ratified by CIRISOntology#3; disposition PROVISIONAL pending CC."*

Small thing, but it is the warrant reach exactly — substance right, authority cited wrong — and this branch is otherwise the cleanest instance of a ruling landing as an enforced gate that I have seen in this ecosystem.

**Accord classification, since the PR body raises it as the consequential half: ruled `axiotic`, unchanged.** Varying it re-ranks outcomes without newly permitting any act; the repairability line does not trip because the record still proves exactly what happened. The chorus-of-traditions design is real but is *provenance* (the warrant axis, provably invisible to the class function), not class. Full reasoning and a naming recommendation for CC's class in CIRISOntology#2.

**Not a blocker for this PR:** CIRISAgent#1007 remains fatal for running any axiotic regime on `default`, and the κ pass (#976) should be measured per axis — content class and warrant separately.

PR #1015, opened 2026-08-08
2.9.12 — one federation identity per node + one handle for trace-sharing consent

### LINKED ISSUE #1004: dream seed Task sets channel_id at top level; context_utils reads only task.context.channel_id → raises
Precise diagnosis from the 2.9.10 backlog cleanse. Supersedes the vaguer #837 framing.

`minimal_dream_processor.py:209` builds the seed Task with `channel_id="dream"` as the **top-level** field and leaves `context=None`.
`context_utils.py:79-89` reads **only** `task.context.channel_id` — never `task.channel_id` — so `hasattr(None, "channel_id")` is False, it falls through, and `:100-102` raises.

One-line fix: populate a `TaskContext` on the seed task. **Zero tests cover this path**, which is why it survived.

Worth a regression test that constructs the seed task and resolves its channel, not just a fix — the failure mode is a dream that cannot start, and nothing currently notices.

### LINKED ISSUE #1009: Accord events are signed with the agent-credits key id, so canonical-server-1 has refused every one for 71 hours
Found in a soak of canonical-server-1 on 2026-08-05. Server-side analysis: `CIRISServer/FSD/RCA_INGEST_REJECTION_2026-08-05.md`.

## What the canonical node sees

```
POST /lens-api/api/v1/accord/events  →  401 verify_unknown_key
  8,631 rejections / 24h, ~6/min, unbroken since 2026-08-02T14:23
  agent-55fe8d181727   4,317
  agent-1ee871dcf31b   4,314
```

persist's diagnostic is precise about why:

```
verify_unknown_key — lookup miss
  envelope_signer_id=agent-55fe8d181727
  accord_public_keys_size=Some(346)
  accord_public_keys_sample=Some(["A1","B1","C1","ciris-agent-bootstrap-25uzoxtlro", …])
```

**Neither key exists in `federation_keys`**, and the shape is wrong for that table. Every registered agent identity is `derive_key_id` output — `ciris-agent-bootstrap-*`. The presented ids are `agent-{hash[:12]}`, which is the **agent-credits** key id (`ciris_engine/schemas/services/agent_credits.py:64`).

So the signing path is using the credits identity namespace against a federation plane that has never heard of it. The server is refusing correctly.

## Consequence

**Trace arrivals stopped at 2026-08-03T23:30 and have not resumed.** Whatever these two agents are producing has reached the canonical node zero times in 71 hours.

If these are the mental-health evaluation runs, none of that data has landed. Worth checking against what you expect to have been delivered — the canonical corpus holds 4,956 trace events, all from agent template `Ally-default-unspecified`, newest 2026-08-03T23:30.

## Fix

Sign accord events with the **federation** key id — the `derive_key_id` value that appears in `federation_keys` — not the credits key id. Those are two identity namespaces and only one of them is admissible on the federation plane.

Worth checking whether any *other* producer path has the same substitution; the two namespaces are both plain strings and nothing in either type system distinguishes them. We are adding a `FederationKeyId` newtype on the Rust side for exactly that reason (CIRISServer#371), but it cannot reach across the language boundary.

## Not your fault alone, and the more useful finding

The server refused correctly 8,631 times a day and **had no instrument that turned this into a signal** — no trace-plane liveness reading, and no reading of a sustained refusal rate from a stable identity set. Filed as CIRISServer#369 and #370. A misconfigured client should not be able to go unnoticed for two days on the strength of everyone behaving correctly.

emooreatx:
## 2.9.12 — route-but-never-root, closed on both axes

Adopts **ciris-server 0.5.161** (persist v30.4.1 + edge v15.21.1) and fixes the agent-side half of why traces were produced perfectly and delivered nowhere — with zero arrivals *and* zero rejections upstream, because nothing was ever attributable.

### 1. One federation identity per node

The Engine minted its own bare Ed25519 + ML-DSA-65 pair under the data dir, because `local_key_path` was the only shape the Python constructor accepted. The node's compose meanwhile resolved its key from the sealed keystore at `<home>/identity`. Two keys — so the self identity-occurrence failed its own scrub-signature and KEX never became authoritative.

`CIRISPersist#616` (in .161) closes it: pass `identity_dir` + `keystore_alias` and the Engine resolves the node's own identity, handing over no key material.

**The alias was the other half, and subtler.** `node_fold` passed `get_edge().signer_key_id()` — the *derived* id `<alias>-<fingerprint>` — as the node's keystore alias, "to align the node to the edge's signer." That alignment was attempted at the wrong layer and was self-feeding: each boot's derived id became the next boot's alias, minting a fresh sealed key. This node had accumulated **61 orphaned `.ed25519.seed.blob` files, one per boot**. The edge derives its id from the Engine's signer anyway, so unifying on the base alias aligns all three by construction.

Also fixes a latent instance of the same class: `_resolve_home()` returned `os.getcwd()` while `get_data_dir()` goes through `get_ciris_home()`. Those agree *only in development mode* — in a managed or Android deployment the node and the Engine would have resolved two different identity dirs. Both now read `get_identity_dir()` / `get_federation_alias()` and cannot drift.

Verified against 0.5.160+'s boot gate, which compares public key bytes — so this is the substrate's own check, not our assertion:

```
gate refusals: 0        sealed-open errors: 0        SignatureInvalid: 0
self_key_id: ciris-agent-bootstrap-dvov7az67u   (engine == edge == compose)
PQC seed adopted from identity/ml_dsa_65.seed
agent reaches WORK in 4.1s; canonical peer ROOTED in ~10s
```

### 2. One handle for trace-sharing consent

Three artifacts must align before a sealed trace is captured, shipped and scored. The writers had drifted into two halves that each authored one:

| path | authored | missing |
|---|---|---|
| setup wizard, Data & Privacy card | capture | **ship** |
| node fold, delivery probe | ship | **capture** |

So the resting state was `capture=True, replication=False` — traces sealed, node healthy, card reporting sharing ON, nothing leaving. `log_federation_consent_drift` was written to shout about precisely this; the detector had been correct and alone the whole time.

Adds `grant_trace_sharing(source)`: one call authors both, returns a typed per-artifact result, never raises. All paths route through it — wizard, data card, legacy env, legacy migration, node fold, delivery probe.

It does not decide for the owner. Session-less callers keep `require_opt_in=True` and fail **closed** on an unreadable signal; they replay an opt-in the owner already gave. Paths carrying the owner's act in-band pass `False`, because the click *is* the consent and the env var may not be written yet.

Two deliberate non-changes, documented at the call site: the DSAR deletion route keeps calling `emit_community_consent_revocation` directly (RECANT obliges a cascade this helper does not run), and the accord adapter keeps its broader `_consent_given` derivation (env OR config graph OR standing CEG grant) — collapsing it would lose the config-graph and post-purge cases.

### Tests

`1467 passed, 2 skipped` across consent, API routes, the accord adapter and runtime; `325` more across persistence and path resolution. New tests lock the invariants that permitted the bug: capture-only never reads as sharing, an unresolvable gate is never a grant, no opt-in path may call a single-artifact emitter directly, and session-less paths may not carry their own read of the opt-in signal.

### What this does *not* yet prove

Traces have not been observed crossing end to end. The identity axis is closed and verified; the consent axis is fixed but was last measured with `live_self_grants: 0` on a run that never claimed the node (`accord_metrics` does no setup, so consent could not be authored regardless). A battery run that claims the node is the outstanding check.

Closes #1009 — the 71-hour canonical rejection. `envelope_signer_id=agent-55fe8d181727`
never matched the node's federation key because the Engine signed as a
locally-minted identity; unifying on `(identity_dir, alias)` is the fix.

Closes #1004

Refs: CIRISServer#380, CIRISPersist#616, CIRISEdge#458

https://claude.ai/code/session_018umjfoyNwpa7BWVmzoayTM


emooreatx:
## End-to-end validated — traces cross, on a node that claims itself

The PR body listed one outstanding check: *"a battery run that claims the node."* It has now run (`safety_battery --live-lens`, en, on this branch against ciris-server 0.5.161). Both axes are closed, and this is the first observation of traces actually leaving.

### Identity axis

```
init failures: 0    gate refusals: 0    sealed-open errors: 0
```

The 0.5.160 gate compares public key bytes, so a clean boot is the substrate's own verdict, not ours.

### Consent axis

```
trace-consent: GRANTED via legacy_migration   (x3)
live_self_grants: 0 -> 1
covers_trace:     false -> true
```

The new handle authored **both** gates. Previously this node sat at `live_self_grants: 0` indefinitely.

### Delivery

```
[DELIVERY-PROBE] canonical ciris-canonical-1-d7bdeu223k ROOTED after ~10s
[DELIVERY-PROBE] canonical ciris-canonical-1-d7bdeu223k KEX PRESENT after ~15s post-root
                 — IdentityOccurrence round synced; replication can now seal envelopes
```

`kex_present` went **false -> true**. It had never been true on this node.

Replication plane — the plane that carries `trace:*`:

| counter | before | now |
|---|---|---|
| `carriage.standing` | `not_exercised` | **moving** |
| `envelopes_served_total` | 0 | **18** |
| `rounds_total` | 0 | **8** |
| `receive.applied_total` | 0 | **4** |

The peer is applying what we send, not merely receiving it.

### One precision

`SHIP CONFIRMED` has not printed, and that is expected rather than a shortfall: that verdict is gated on the APPLICATION plane's `envelopes_sent_total`, and the diagnostics themselves note that anti-entropy replication — the plane carrying `trace:*` — increments none of those counters (CIRISEdge#434). The replication-plane standing above is the authoritative signal for trace carriage.

So: `knows_peer: true, kex_present: true, deliverable: true` — the inverse of the state CIRISServer#380 and CIRISEdge#458 each produced on their own.

emooreatx:
## Correction to the previous comment — traces did NOT cross

My previous comment claimed traces shipped end-to-end. **That was wrong**, and CI caught what I did not.

`sealed_and_persisted = 0` for that run. **Zero traces were sealed**, so none could cross. The 18 replication envelopes were other kinds (identity occurrences, consent artifacts) — the replication plane was genuinely moving, but not carrying `trace:*`. I read plane-level carriage counters as trace carriage; they are not the same thing, and the run log had `synchronous sign_ed25519` errors 9 times that I did not check.

### What still stands

These were independently verified and are unaffected:

- **Identity unified** — 0 gate refusals against 0.5.160's public-key comparison, one `self_key_id` across engine/edge/compose
- **KEX** — `false -> true`, PRESENT ~15s post-root. Genuinely new.
- **Consent** — `live_self_grants: 0 -> 1`, `covers_trace: false -> true`, via the new handle

### What is actually broken, and it is a real incompatibility

Adopting keystore custody makes the classical signer **async**, so persist refuses the synchronous verb:

```
RuntimeError: classical (Ed25519) sign: synchronous sign_ed25519 is unavailable
for a hardware-custodied classical key (its HardwareSigner::sign is async);
use sign_hybrid
```

Two consumers call it. I fixed ours — `audit_service` now uses `local_sign_hybrid()["classical_sig"]` (same 64-byte Ed25519 signature, unchanged wire shape), which cleared the audit cascade completely: `Failed to add to persist audit chain` and `Hash chain data not generated` both went 13/36 -> **0/0**.

The other consumer is **lens-core**, and it is not ours to fix. Its documented cohabitation contract is:

```
The cohabitation path calls these engine Python methods (the seal is HYBRID):
  - engine.local_sign(canonical_bytes: bytes)          -> 64-byte Ed25519 sig
  - engine.local_pqc_sign(canonical || ed25519_sig)    -> ML-DSA-65 sig
```

`local_sign` is the verb keystore custody removes. So every `capture_event` fails with `capture DEGRADED`, nothing seals, and the three `accord_metrics` QA tests fail exactly as reported: *"No traces captured — accord_metrics adapter may not be loaded."* The adapter is loaded; it cannot sign.

**This means CIRISPersist#616's remedy and lens-core's seal contract are mutually exclusive today.** A host that adopts one identity (as CIRISServer#380 requires) loses trace sealing; a host that keeps sealing keeps two identities. That is the ask to file upstream: lens-core's cohabitation seal should call `local_sign_hybrid` — persist's single hybrid-sign verb (v17.7.0 / CIRISPersist#470) — exactly as our audit service now does.

This PR is **blocked on that** for the trace half. The identity, consent, audit-signing and dream fixes stand on their own.

PR #1032, opened 2026-08-14
fix(ci): safety evidence sync has been red for 7 days — main now requires a PR

emooreatx:
## The failure

`Safety evidence sync` has failed **every scheduled run since 2026-08-08** — seven consecutive days:

```
remote: error: GH013: Repository rule violations found for refs/heads/main
remote: - Changes must be made through a pull request.
! [remote rejected]     main -> main (push declined due to repository rule violations)
```

A repository ruleset now requires pull requests for `main` — the same rule that made the 2.9.14 merge need `--admin`.

## Why it matters more than a red check

Nothing about the harvest was broken. `tools/harvest_safety_evidence.py` ran fine every day; only the **delivery** was illegal, so the job died at the last step with all the work done.

The workflow exists because Actions artifacts expire at 90 days and safety-battery evidence was days from the cliff when it was written. For the past week that copy has not been made, while the artifacts kept aging — the exact loss this job guards against, arriving by a route it did not cover.

## The fix

Push to a long-lived `evidence/safety-sync` branch and open (or refresh) a PR, instead of pushing to `main`.

The retention guarantee holds at **push** time, not merge time: once bundles are on the remote they are off the artifact clock, so a PR waiting on a human costs nothing that matters.

- Force-pushed each run, so it is one accumulating PR rather than a new one per day. Safe because nothing else writes that branch and its contents are regenerated from artifacts every run.
- Each run rebases onto current `main` (checkout is `ref: main`, then `checkout -B`), so the PR never goes stale.
- Added `pull-requests: write`; the job only had `contents: write`.

## Not doing

**Not** adding a ruleset bypass for the bot. The rule is doing its job; the workflow was the thing out of step.

**Not** bumping the version — CI-only, and 2.9.14 was tagged minutes ago, so a bump would immediately desync `main`'s version string from the release tag for no runtime change.

## Verification

YAML parses; 6 steps resolve in order. No `push origin main` remains in the file outside the explanatory comment. The first scheduled run after merge (11:40 UTC) is the real test — or trigger `workflow_dispatch` to check sooner.

### DIFF
--- a/.github/workflows/safety-evidence-sync.yml
+++ b/.github/workflows/safety-evidence-sync.yml
@@ -51,6 +51,8 @@ concurrency:
 permissions:
   contents: write
   actions: read               # listing + downloading artifacts
+  pull-requests: write        # main is ruleset-protected: evidence lands via PR,
+                              # not a direct push (see the delivery step)
 
 jobs:
   sync:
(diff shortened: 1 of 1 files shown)


emooreatx:
Superseded by #1035, which bundles this with the other two and is merged as 7788e66e6. No version bump — nothing in the agent changed.

PR #1034, opened 2026-08-14
fix(docs): results page shipped with its script dead — duplicate const kills the whole page

emooreatx:
## The page is not partly broken — it is entirely broken

https://cirisai.github.io/CIRISAgent/ shows `loading…` because its script does not parse:

```
SyntaxError: Identifier 'esc' has already been declared
```

Adding the questions/rubrics tab introduced a second `const esc=` alongside the one already declared at line 223. Duplicate `const` at the same scope is a **parse** error, and JavaScript aborts the entire script rather than the offending statement — so every tab is dead, not just the new one.

## Why nothing caught it

Every signal that could plausibly have flagged this was green:

| check | result |
|---|---|
| page loads | `200` |
| deployed bytes vs repo | byte-identical |
| `data/batteries/index.json` | `200` |
| `data/batteries/en.json` | `200` |
| all 29 `data/rubrics/*.json` | `200` |

Nothing asked whether the code could **run**. And the symptom actively misleads: `loading…` looks like a failed fetch, so the natural first move is to inspect the data — which was correct the whole time.

## The fix

Delete the duplicate; use the existing declaration. The two were functionally identical (`s??""` vs `t==null?"":t`).

Verified **sufficient**, not merely necessary — replaying the page's own render expressions against the live JSON:

```
meta:      en_mental_health_v4 · battery v4 · 9 questions
questions: 9 rendered
criteria:  7 rendered — U1 U2 U3 U4 U5 U6 U7
```

## The guard

`tests/docs/test_results_page_script_parses.py`, with two deliberately overlapping checks:

1. `node --check` — the real parser, where node exists.
2. A duplicate top-level `const`/`let` scan in pure Python, which **always** runs.

The second exists because a gate that silently skips when a tool is missing is no gate at all on the machine that lacks it — and this defect's whole character was passing every check that did run.

Both were confirmed to fail on the reintroduced bug and pass again after restoring, so they lock the fix rather than merely accompanying it.

### DIFF
--- a/docs/results/index.html
+++ b/docs/results/index.html
@@ -377,7 +377,6 @@ const LANG_NAMES={am:"Amharic",ar:"Arabic",bn:"Bengali",de:"German",en:"English"
   ta:"Tamil",te:"Telugu",th:"Thai",tr:"Turkish",uk:"Ukrainian",ur:"Urdu",vi:"Vietnamese",
   yo:"Yoruba",zh:"Chinese"};
 const RTL=new Set(["ar","fa","ur"]);
-const esc=t=>String(t==null?"":t).replace(/[&<>"]/g,c=>({"&":"&amp;","<":"&lt;",">":"&gt;",'"':"&quot;"}[c]));
 
 J("data/batteries/index.json").then(d=>{
   const sel=$("#how-lang");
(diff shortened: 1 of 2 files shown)


emooreatx:
Superseded by #1035, which bundles this with the other two and is merged as 7788e66e6. No version bump — nothing in the agent changed.

PR #1040, opened 2026-08-15
fix(setup): an OAuth user is asked for a password they never set

emooreatx:
Reported from the desktop wizard, mid-setup:

```
HTTP 400 {"detail":"New user password must be at least 8 characters"}
```

**The server is right.** `/v1/setup/complete` computes `is_oauth_user = bool(oauth_provider)`; when true it generates a random password for an account that will never use one, and when false it requires a real one. The wizard was telling it this was a **password account**.

## One question, two answers

The two completion branches disagreed about what makes a user an OAuth user:

| branch | test |
|---|---|
| CIRIS-proxy | `oauth_provider = currentState.oauthProvider` |
| BYOK | `isGoogleAuth && currentState.googleUserId != null` |

So the same signed-in user is an OAuth account on one path and a password account on the other. The reporter was on **BYOK** (own API key), where an OAuth sign-in whose provider returned no subject id fell through to the password branch — and was asked for a password that does not exist.

`googleUserId` is evidence about **which account**, not about **whether OAuth happened**. It must not gate this.

## The fix

`SetupFormState.isExternalAuth` and `SetupFormState.effectiveOAuthProvider` are now the single derivation both branches read, so the flag and the value it is paired with cannot disagree — and the two paths cannot drift apart again.

## A second, latent bug this surfaced

`isGoogleAuth` is **misnamed**: it is set from `isAuth` for both Google *and* Apple. Any check written against "Google" literally sends Apple users down the password path. The field now documents this, and `apple_is_oauth_too` covers it.

## Tests

`SetupExternalAuthTest` — **4/4**, pinning the derivation rather than either call site (verified from the JUnit XML, `tests="4" skipped="0" failures="0"`, since a task that runs nothing also reports BUILD SUCCESSFUL):

- OAuth with a null subject id is still OAuth — *the reported case*
- Apple is OAuth too
- HA add-on is external auth but not OAuth (SUPERVISOR_TOKEN, no password either)
- a genuine password account is **not** external — or the server would silently generate a random password and discard the one the user chose

## Provenance

Found while adopting this client into CIRISServer, which vendors it. The identical fix is landing there; this is the upstream half so the two trees don't diverge on it.

### DIFF
--- a/client/shared/src/commonMain/kotlin/ai/ciris/mobile/shared/viewmodels/SetupState.kt
+++ b/client/shared/src/commonMain/kotlin/ai/ciris/mobile/shared/viewmodels/SetupState.kt
@@ -455,6 +455,45 @@ data class SetupFormState(
     val submissionError: String? = null
 ) {
     /**
+     * **Did this user authenticate through an external provider?** — the ONE
+     * derivation, because answering it twice is what broke setup.
+     *
+     * `/v1/setup/complete` computes `is_oauth_user = bool(oauth_provider)` and,
+     * when true, GENERATES a random password for an account that will never use
+     * one. When false it refuses with *"New user password must be at least 8
+     * characters"* — correct for a password account, and nonsense to someone who
+     * has just signed in with Google.
+     *
+     * The two call sites disagreed. The CIRIS-proxy branch passed
+     * [oauthProvider] straight through; the BYOK branch required
+     * `isGoogleAuth && googleUserId != null`. So a BYOK user who signed in with
+     * OAuth, but whose provider returned no subject id, read as a PASSWORD user
+     * and was asked mid-wizard for a password they had never set.
+     *
+     * [googleUserId] is evidence about WHICH account, not about WHETHER OAuth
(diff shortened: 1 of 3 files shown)


emooreatx:
Superseded — closing rather than merging.

This fix (`isExternalAuth` / `effectiveOAuthProvider`, and `SetupExternalAuthTest`)
already landed on `main` in **2.9.17** (`0b72ee5b7`), so there is nothing here left
to deliver.

Merging would also be actively destructive: the branch is **316 commits behind**
`main`, so the merge would revert roughly **37,500 lines across 114 files** — every
release since. CI is green on it only because it is green against a tree from four
days ago.

The behaviour it fixed is verified present on `main`:

```
SetupState.kt:679      val isExternalAuth: Boolean
SetupState.kt:687      val effectiveOAuthProvider: String?
SetupViewModel.kt:2461 val isExternalAuthUser = currentState.isExternalAuth
SetupViewModel.kt:2487 admin_password = if (isExternalAuthUser) null else currentState.userPassword
```

PR #1055, opened 2026-08-16
fix(client): iOS/macOS release prep — claim-PIN, CIRIS_HOME, verify 405, OAuth/setup UX, version align

emooreatx:
Release-prep fixes surfaced while driving the desktop and iPhone builds end to end for the App Store upload.

## Setup / claim
- **Claim PIN read from the durable file.** The node mints a one-time PIN into `0600 <CIRIS_HOME>/claim_pin`, but the client only latched it from the node's live stdout banner — which iOS never owns and which is racy on desktop. Added `PythonRuntimeProtocol.readLocalClaimPin()`, a race-free on-demand read at claim time. Local node was left **UNCLAIMED** before this.
- **`get_config_paths()` honors `CIRIS_HOME`.** First-run detection READ `~/ciris` while the wizard WROTE under `get_ciris_home()`, wedging desktop in a first-run loop. Read path and write path now agree (+ regression test).

## Verify 405 (server half was fixed, client half wasn't)
- The client polls verify status every 30s. iOS is an **agent build** (`isNode=false`), so it kept GETting the POST-only `/v1/auth/attestation` → **HTTP 405, 132×/session**, verify never loaded. Now uses the proxied `/v1/system/verify-status` in agent mode too (node serves it on :4243, agent proxies on :8080). **Device-confirmed: 405 count 0 (was 132).**

## Login / setup UX
- **First-run shows the Login screen on desktop** (was auto-skipped straight to local setup because `googleSignInCallback == null` was misread as "no OAuth" — untrue since the browser-handoff flow landed).
- **Grey out the OAuth button when no credentials are wired** — probe the node's `/v1/auth/oauth/providers` (a provider appears only once the client_secret CI bakes from GH secrets is present). Native Apple sign-in (iOS) stays live.
- **YOU step reordered** to Age → Account → Federation identity. Age moved to the top and made **required** (it seeds the fed-ID and gates minor-stewardship). Local username/password render only for non-OAuth; the fed-ID label auto-populates from the username / OAuth id (overridable). `admin` rejected client-side.

## Version alignment — closes the `.173/.176` banner
- `main` shipped misaligned: requirements pinned `ciris-server 0.5.176` but `CLIENT_VERSION`, the Android gradle pin, and the **iOS bundled substrate** were `0.5.173` → "Node is v0.5.173 — app built for v0.5.176". Bumped the client + Android gradle pin, refreshed the Android wheels **and** the iOS Resources.zip substrate to 0.5.176. **Device-confirmed: node `/v1/health` now reports 0.5.176, banner gone.** `check_version_alignment.py` passes.

## QA runner
- iOS physical test now drives setup → self-claim **and** exchanges a real message with the agent (login → wait WORK → `/v1/agent/interact`), degraded-LLM aware. Added a macOS desktop E2E driver.

## Validation
- **iOS device (fresh install):** local-login setup → self-claim (`owner=ciris-founder-node-…`) → verify **405 gone** → node reports **0.5.176 (banner cleared)**.
- **macOS:** backend boots on 0.5.176 (local pip substrate was stale at 0.5.134 — `Engine` rejected `identity_dir`; upgraded); `CIRIS_HOME` resolves first; Login + Google button render; wizard drives. Self-claim on macOS is gated only by a keychain-persisted owner from prior test runs — the claim path itself is proven clean on iOS (uninstall wipes the keychain).
- Shared Kotlin compiles (iOS + desktop); localization + first_run tests green.


emooreatx:
Landed in 2.9.22 via #1054 (482b45f39), not merged directly.

This branch was merged into `fix/2.9.22-windows-hardening` so the two .22 workstreams shipped as one release, and #1054 was squash-merged — so these commits are in `main` but this PR does not show as merged. Verified on main: `client/iosApp/substrate.lock.json` reads server 0.5.176 with the refreshed binaries, and `check_version_alignment.py` passes.

Worth recording that this PR independently closed two things the Windows workstream was stuck on:

- **compose-gate.** It had blocked #1054 all day on the iOS lock saying 0.5.173 against a 0.5.176 pin. I could not fix it without macOS — the bundle carries no `ciris_server` dist-info, so the lock really is the only thing a non-macOS runner can check against, and hand-editing it would have defeated the gate. This PR carried the actual refreshed binaries *and* the lock, which is the correct fix.
- **A login divergence I had chased and could not explain.** Windows rendered a provider chooser where Linux rendered the credential form directly; I made the harness handle both and flagged it as an open product question. This PR answers it: first run auto-skipped the Login screen because `googleSignInCallback == null` was read as "no OAuth configured", which stopped being true when the browser-handoff flow landed. Never a platform difference.

Also: `get_config_paths()` honouring CIRIS_HOME here is the READ half of the split-home bug #1054 fixed on the write/harness half — independent discovery, same root cause.

One follow-up left open from this work: the iOS refresh workflow only rewrites `substrate.lock.json` for libs it actually refreshes, so a no-op refresh leaves a stale lock behind (it reported "no change" twice while the lock was wrong).

PR #1112, opened 2026-08-23
fix(desktop): the token-refresh handshake wrote a name the agent never read (2.9.36)

emooreatx:
Diagnosed from a user's log. Their desktop agent works for about an hour after setup, then stops answering — permanently.

```
[BILLING_PROVIDER] Started:  auth_mode: JWT (Google ID token)
                             has_refresh_callback: False
ciris_primary error: 401 - {'message': 'Invalid authentication token'}
[INTERACT_TIMEOUT] message_id=… timed out after 110s without an agent response
```

## Why an hour

The CIRIS-hosted proxy is authenticated with the OAuth ID token itself (`api_key=id_token`), and those expire in about an hour. There is a handshake to replace one:

1. agent gets a 401 → writes `.token_refresh_needed`
2. client polls it → silently re-signs in
3. client rewrites `.env` with the fresh token
4. client writes `.config_reload` → agent reloads and swaps its key

Every step ran correctly on desktop and **achieved nothing**:

| platform | writes | Python readers |
|---|---|---|
| Android | `CIRIS_BILLING_GOOGLE_ID_TOKEN` | 19 ✓ |
| iOS | `CIRIS_BILLING_APPLE_ID_TOKEN` | 14 ✓ |
| **Desktop** | **`CIRIS_BILLING_OAUTH_TOKEN`** | **0** ✗ |

So the agent re-read the same expired token, logged `API key unchanged after refresh`, reset the circuit breaker, and 401'd again. Forever. Nothing in the log named the cause — the timeline even looks like a working credential, since the *billing* endpoint accepted the same identity seven seconds before the LLM proxy rejected it.

## The fix

- Desktop writes the Google name, and clears the dead legacy key out of `.env`.
- The agent reads all three names, so a client and an agent on different versions still complete the handshake.
- Selection is by **JWT `exp`, not name order**. A real `.env` carries the setup-time token *and* the refreshed one, and first-non-empty returns the expired one. It also never re-selects the value already in hand when a different copy exists — "keep the credential that is currently 401ing" is not a resolution of that tie.

## DRY, because the drift *was* the bug

The handshake's directory, its two filenames and its token variables now live in one module per language — `ciris_engine/logic/utils/token_handshake.py` and `TokenHandshake` in commonMain. Removed: a second copy of the refresh request in the billing provider, a third spelling of `.config_reload` in the setup route, three sets of private constants in the platform updaters.

## Logging, because the outage was silent

One `[TOKEN_HANDSHAKE]` prefix across both processes so the conversation greps as one story. The request names the path it wrote and why — the only way to see that the two halves are watching different directories. The no-op path now says which variables it consulted, separating "the client never wrote a new token" from "the token is new and still rejected".

## Tests

12 cases reproducing the chain and pinning the contract: the names the two languages must agree on, that no site re-spells a filename (a sweep over `ciris_engine/`), freshest-wins selection with real JWT shapes, the tie-break, and that the ask logs its path. Desktop Kotlin compiles clean.

https://claude.ai/code/session_01EVMJwukQjscjwLQz9NZktK

### DIFF
--- a/ciris_adapters/ciris_hosted_tools/services.py
+++ b/ciris_adapters/ciris_hosted_tools/services.py
@@ -173,10 +173,16 @@ class CIRISHostedToolService:
         Returns:
             Google ID token if available, None otherwise
         """
-        # First try environment (direct or loaded at startup)
-        token = os.environ.get("CIRIS_BILLING_GOOGLE_ID_TOKEN")
+        # Same selector as billing and the LLM services. Reading only the
+        # Google name here meant that after a desktop client refreshed under a
+        # different variable, hosted tools and tool-balance requests went on
+        # sending the stale setup token — or none — while the rest of the agent
+        # had already recovered.
+        from ciris_engine.logic.utils.token_handshake import read_proxy_token
+
+        token, source_var = read_proxy_token()
         if token:
-            logger.info(f"[HOSTED_TOOLS] Got token from env CIRIS_BILLING_GOOGLE_ID_TOKEN (len={len(token)})")
+            logger.info(f"[HOSTED_TOOLS] Got token from env {source_var} (len={len(token)})")
         else:
             token = os.environ.get("GOOGLE_ID_TOKEN")
             if token:
--- a/ciris_adapters/ciris_hosted_tools/services.py
+++ b/ciris_adapters/ciris_hosted_tools/services.py
@@ -216,9 +216,20 @@ class CIRISHostedToolService:
                     logger.info(f"[HOSTED_TOOLS] Checking {env_path} exists={env_path.exists()}")
                     if env_path.exists():
                         with open(env_path) as f:
+                            # SAME NAMES AS THE SELECTOR. Scanning only for
+                            # CIRIS_BILLING_GOOGLE_ID_TOKEN here is the same
+                            # drift one level deeper: a desktop client that
+                            # refreshed under a different variable wrote it
+                            # into this very file, and this loop would walk
+                            # straight past it.
+                            from ciris_engine.logic.utils.token_handshake import PROXY_TOKEN_VARS
+
+                            found: dict[str, str] = {}
                             for line in f:
                                 line = line.strip()
-                                if line.startswith("CIRIS_BILLING_GOOGLE_ID_TOKEN="):
+                                for var in PROXY_TOKEN_VARS:
+                                    if not line.startswith(f"{var}="):
+                                        continue
                                     # Handle quoted and unquoted values
                                     value = line.split("=", 1)[1].strip()
                                     if value.startswith('"') and value.endswith('"'):
@@ -226,11 +237,20 @@ class CIRISHostedToolService:
                                     elif value.startswith("'") and value.endswith("'"):
                                         value = value[1:-1]
                                     if value:
-                                        token = value
-                                        logger.info(
-                                            f"[HOSTED_TOOLS] ✓ Loaded fresh token from {env_path} (len={len(value)})"
-                                        )
-                                        break
+                                        found[var] = value
+                            if found:
+                                # Rank them the same way the selector does, so
+                                # the file path and the environment path cannot
+                                # disagree about which token is current.
+                                from ciris_engine.logic.utils.token_handshake import jwt_expiry_epoch
+
+                                var, value = max(
+                                    found.items(), key=lambda kv: (jwt_expiry_epoch(kv[1]) or 0.0)
+                                )
+                                token = value
+                                logger.info(
+                                    f"[HOSTED_TOOLS] Loaded fresh token from {env_path} under {var} (len={len(value)})"
+                                )
                         if token:
                             break
                         else:
--- a/ciris_adapters/ciris_hosted_tools/services.py
+++ b/ciris_adapters/ciris_hosted_tools/services.py
@@ -239,14 +239,19 @@ class CIRISHostedToolService:
                                     if value:
                                         found[var] = value
                             if found:
-                                # Rank them the same way the selector does, so
-                                # the file path and the environment path cannot
-                                # disagree about which token is current.
-                                from ciris_engine.logic.utils.token_handshake import jwt_expiry_epoch
+                                # THE SAME RANKING, not a second one that
+                                # happens to agree on the easy cases. Comparing
+                                # expiry alone here left the opaque tie to
+                                # dictionary insertion order — Google first —
+                                # while the shared selector resolves that same
+                                # tie toward the legacy-desktop name. The file
+                                # path and the environment path would then pick
+                                # different tokens from identical inputs, so
+                                # hosted tools sent the stale credential while
+                                # billing and the LLM used the refreshed one.
+                                from ciris_engine.logic.utils.token_handshake import rank_candidates
 
-                                var, value = max(
-                                    found.items(), key=lambda kv: (jwt_expiry_epoch(kv[1]) or 0.0)
-                                )
+                                var, value = rank_candidates(list(found.items()))[0]
                                 token = value
                                 logger.info(
                                     f"[HOSTED_TOOLS] Loaded fresh token from {env_path} under {var} (len={len(value)})"
--- a/ciris_engine/constants.py
+++ b/ciris_engine/constants.py
@@ -6,11 +6,11 @@ from typing import List
 from ciris_engine.schemas.runtime.canonical_peer import CanonicalBootstrapPeer
 
 # Version information
-CIRIS_VERSION = "2.9.35-stable"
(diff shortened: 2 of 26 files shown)


emooreatx:
Thanks — three of the four were real defects in the fix, and one of them was in my own regression test. All addressed in 28c784622.

**P1 · billing kept its own reader** — right, and it would have moved the outage rather than ended it: the LLM key recovers while billing goes on sending the stale setup-time token, answers AUTH_EXPIRED, and gates every interaction. Both read paths (`_get_current_token()` and `create_billing_token_handler`) now go through `read_proxy_token()`.

**P1 · the tie-break could downgrade** — correct, and this one was mine to catch. "Prefer any copy that differs" is right when both tokens are opaque and ours is the one being refused, but wrong when ours is genuinely the freshest: a stale sibling in `.env` would be selected on every reload, including reloads the setup route triggers for unrelated reasons. An alternative must now be at least as fresh (`>=` on the parsed `exp`), which keeps the legacy-opaque case working (both score 0) while refusing the downgrade. Both directions are now tested.

**P1 · the runtime-wide handler** — the replica mechanism named isn't in the tree (`CIRIS_LLM_REPLICA_COUNT` has no references), but the underlying defect is real and worse than described: `create_llm_token_handler` read `OPENAI_API_KEY` and pushed it through `update_llm_services_token`, which only ever touches CIRIS-proxy services — and those authenticate with the ID token. So on hosted proxy (variable empty) it refreshed nothing, and with a BYOK leftover in that variable it would push someone else's provider key over a good token. It reads the proxy token now; the two tests that pinned the old behaviour are rewritten, plus one asserting an unrelated `OPENAI_API_KEY` is ignored.

**P2 · the contract test scanned comments** — entirely right, and embarrassing: the explanatory comment above the assignment names both correct variables, so reverting `tokenKey` would have left it green and let the original outage back in. It now strips comments and asserts the assignment itself. Verified by reintroducing the bug — two tests fail, and pass again on restore.

19 tests on the handshake, 491 in the surrounding suites.

emooreatx:
Right again, and this one mattered: routing billing's *reads* through the shared selector wasn't enough, because `_get_current_token()` consulted the injected callback **first** and returned the moment it produced anything different.

So with a stale `CIRIS_BILLING_GOOGLE_ID_TOKEN` beside a fresh `CIRIS_BILLING_OAUTH_TOKEN`, the sequence was exactly as you describe: the handler installs the fresh token, the next billing request calls `get_fresh_token()` (still first-non-empty over Google/Apple at billing_helpers:129-130), that returns the **stale** value, `_get_current_token()` sees it differs from what's installed, writes it back, and returns before ever reaching `_read_proxy_token()`. The refresh was undone one request later and AUTH_EXPIRED resumed.

Fixed in 475859f9e:

- **The selector decides, and it goes first**, with the currently-held token passed in so the opaque-token tie-break works (neither side has a readable `exp` in the legacy case).
- **The callback is demoted to what it actually is** — a way for a caller to supply a token from somewhere other than `.env` — and is consulted only when the environment offers nothing. It also uses the selector now, so it can't re-introduce the choice.

Four new tests: a stale callback cannot reinstall an expired token; the same with opaque tokens; the callback still serves an empty environment (demoted, not removed); and the factory doesn't re-implement the selection.

The ordering change also surfaced three tests that depended on environment state by luck rather than declaring it — the two resource_monitor callback tests (which exercise the fallback branch and must say so), and `test_env_override_ciris_billing_url`, which tests the lower-precedence variable while a leaked `CIRIS_BILLING_API_URL` silently wins. All three now state their environment.

**496 passed, 0 failed** across the billing/token/llm/resource-monitor suites — including that last test, which was already red before this branch started.

emooreatx:
All three addressed in a785e7660 — the alternation one is the sharpest catch on this PR.

**Opaque tokens alternated on every request.** Exactly as described: breaking a tie by "prefer whichever copy is not the one in hand" isn't a rule, it's a coin that remembers. Holding A it returned B; holding B, the stable sort put A first again, the guard no longer fired, and it returned A — every billing request swapping credentials, one of the two always the rejected one, in precisely the compatibility path the new tests exercise.

The answer now depends **only on the candidates**, never on what is installed: readable `exp` first, then a documented order for tokens whose expiry cannot be read at all. `current` survives only as a log detail. That also retires your earlier downgrade finding by construction — a selector that cannot see what you hold cannot walk you backwards off it. Tested by asserting the answer is identical across `current ∈ {"", A, B}`, and that five consecutive `_get_current_token()` calls settle on one value.

**The callback was skipped whenever the environment held anything.** Right, and I'd created the mirror image of the bug I was fixing: callback-first let a stale value overwrite a fresh one, callback-never silenced callers fetching from secure storage — which is what the parameter is documented for. It's now a candidate, ranked with the rest. One refinement your case implies: candidates are deduped by value, so the legacy first-non-empty callback echoing a variable already in `.env` gains no precedence over it. Echoing isn't sourcing.

**Hosted tools still read the names itself.** Confirmed — `_get_google_id_token()` looked only at the Google and legacy variables, so tool and tool-balance requests kept using the stale setup token after billing and the LLM had recovered. It uses the shared selector now; that was the fourth and last consumer.

Also raised new-module coverage 82.5% → 97.5% (the three paths, the `CIRIS_HOME` fallback including a resolver returning a `str`, directory creation, an unwritable home reported rather than raised, the nothing-anywhere log, and the unreadable-JWT cases), since the PR gate wanted 80% on new code.

**507 passed** across billing / token / llm / resource-monitor / hosted-tools.

emooreatx:
All three fixed in c06f77eb4 — and grepping for the pattern rather than only the named sites turned up five more.

**The lazy initializer (P1)** — correct, and the symptom is worse than a missing provider: with a legacy-desktop token present it returned `None`, so `/credits` answered with the unlimited-credit response and the purchase routes reported nothing configured, all while a valid token sat in the environment. Compatibility shouldn't depend on a provider already existing, so it uses the selector now.

**The tools route** — same fix; a header still wins, since that's the caller naming the identity.

**Restarts regressed** — this was the sharpest of the three. The proxy has no provider enum of its own (it rides `OPENAI_COMPATIBLE`), so its startup credential came from `OPENAI_API_KEY` — exactly where the setup wizard stashes the hour-lived ID token, once. Every restart therefore rebuilt the primary service around the setup-time value no matter how many refreshes had replaced it since. Rather than teaching desktop to also write `OPENAI_API_KEY` (an ID token isn't a provider API key), proxy startup now reads the handshake variables when `is_ciris_proxy_url()` is true.

**Five more the sweep found:** two in `service_initializer` (mobile JWT path, secondary service), one in `billing_helpers`, one more in the billing routes, and the two I'd have most regretted missing — the capability gates in `platform_detection.py` and `api/adapter.py`. Those decide whether hosted-proxy features exist at all, and they asked only about the Google name, so a client that refreshed under a different one looked exactly like a user who had never signed in: the capability withdrawn with a valid token in hand. They take a new `has_proxy_token()`.

Zero modules outside `token_handshake.py` now read a token variable directly — the setup wizard still *writes* the canonical name, which is its job.

Since every review round has found another quiet copy of that list, the line is now held by a test: one grep over `ciris_engine/` and `ciris_adapters/` that fails on any direct read, plus its inverse asserting the seven known consumers import the selector.

**804 passed** across billing / token / resource-monitor / hosted-tools / platform-detection / service-initializer.

emooreatx:
Both real, and both in the ranking itself rather than a missed consumer — fixed in cd25f3973.

**Expired JWTs outranked every opaque token.** The key was `(exp or 0.0, …)`, and an expired token still carries a large positive epoch while a token whose format has no `exp` scores zero — so a credential we can *prove* is dead beat one that might be alive, including the `google:{user_id}` shape a callback legitimately returns. The callback could fetch a good token and be ignored in favour of the exact value we were being refused for.

Expiry status is a tier now, with ordering inside each:

| tier | | within |
|---|---|---|
| 2 | live — `exp` in the future | later expiry first |
| 1 | unknown — no readable `exp` | the documented opaque order |
| 0 | expired — provably dead | least-stale first |

Expired candidates stay as a last resort rather than being dropped: sending a stale token and getting a clean 401 is more useful than sending no credential at all.

**The `.env` scan had its own ranking** — `max` on expiry alone, so the opaque tie fell to dictionary insertion order (Google first) while the shared selector resolves it toward the legacy-desktop name. Identical inputs, different answers, which is the exact disagreement this PR exists to end. The ordering is now `rank_candidates()`, both paths call it, and a test asserts the file path and the environment path return the same token from the same pair.

Five more tests on the tiering (expired-loses-to-opaque from both the callback and the environment, live-still-beats-opaque, expired-still-offered-when-it-is-all-there-is, least-stale-wins-among-expired) and two on the file/environment agreement.

**811 passed.**

PR #1114, opened 2026-08-24
fix(qa): the trace-ship harness reported a healthy transport as broken

emooreatx:
Found by running the trace-ship harness live against the canonical peer on 2.9.37 / ciris-server 0.5.188. The product finding is filed as [CIRISServer#487](https://github.com/CIRISAI/CIRISServer/issues/487); this PR is the harness half.

## The verdict contradicted itself

```
transport-rooted : ❓ probe verdict not found  (edge.knows_peer — authoritative)
❌ Canonical NOT transport-rooted (edge.knows_peer=false) — neither rounds nor
   trace envelopes can flow. Check the canonical boot-prime + bootstrap dial.
delivery_status  : phase=kex-present-await-ship started=True edge_up=True
→ deliverable=true but trace not landing
```

The node log for that same run:

| measure | value |
|---|---|
| canonical | rooted from baked KeyRecord |
| anti-entropy rounds | 300 |
| timeouts | 30 |
| inbound envelopes | 50 |
| **FramesDropped** | **0** |

Five other envelope kinds were replicating happily. The transport was fine — the `[DELIVERY-PROBE]` line simply never landed, and an **absent** verdict was rendered as `false`. That points the reader at the boot-prime and the bootstrap dial, both healthy, and away from the ship stage, which is the part that is genuinely stuck.

- Unknown is now its own outcome; the hard `knows_peer=false` verdict is reachable only when the probe actually said `False`.
- When the probe line is missing, the verdict is taken from `delivery_status()` — the same per-peer `knows_peer` / `kex_present` / `deliverable` the node itself acts on, read as data rather than prose.

## A check that cannot run is not a check that failed

The Lens public-keys endpoint is retired upstream and says so in its own 404 body — `lens retired; see /v1/identity and /lens/api/v1/*`. Two checks failed red for it. Red that doesn't mean anything is how a suite teaches people to ignore red, and this run had a third red that *did* matter.

Retirement is now detected from the declared body and reported as **SKIP**, a third outcome distinct from pass and fail. An ordinary 404, or a 500 that merely contains the word "retired", still fails.

## A failure that says nothing

```
❌ FAIL Verb Second Pass Trace:
```

Nothing after the colon: the only failure path was `return False, str(e)`, and `str(TimeoutError())` is the empty string. Failures now name the exception type and report how far the test got; the generic loop does the same and logs the traceback.

## Tests

Replay the real shapes — the `[DELIVERY-STATUS]` line verbatim from the run that misreported, peer selection by canonical target, unreadable logs and a reader that raises, the retirement predicate's true and false cases, plus source-level guards that the hard verdict stays behind `is False` and that both Lens checks route retirement to SKIP.

https://claude.ai/code/session_01EVMJwukQjscjwLQz9NZktK

### DIFF
--- a/ciris_engine/constants.py
+++ b/ciris_engine/constants.py
@@ -6,11 +6,11 @@ from typing import List
 from ciris_engine.schemas.runtime.canonical_peer import CanonicalBootstrapPeer
 
 # Version information
-CIRIS_VERSION = "2.9.37-stable"
+CIRIS_VERSION = "2.9.38-stable"
 ACCORD_VERSION = "1.2-Beta"
 CIRIS_VERSION_MAJOR = 2
 CIRIS_VERSION_MINOR = 9
-CIRIS_VERSION_PATCH = 37
+CIRIS_VERSION_PATCH = 38
 CIRIS_VERSION_BUILD = 0
 CIRIS_VERSION_STAGE = "stable"
 CIRIS_CODENAME = "Context Engineering"  # Codename for this release
--- a/ciris_engine/logic/runtime/edge_runtime.py
+++ b/ciris_engine/logic/runtime/edge_runtime.py
@@ -322,6 +322,45 @@ def _spawn_delivery_rooting_probe(engine: Any, edge: Any) -> None:
         except Exception as exc:  # noqa: BLE001
             logger.debug("[DELIVERY-STATUS] phase=%s accessor error (non-fatal): %s", phase, exc)
 
+    def _log_trace_plane(phase: str) -> None:
+        """Surface `ciris_server.node_state().trace_plane` as a [TRACE-PLANE] line.
+
+        SAME REASON AS _log_delivery_status ABOVE: the accessor is in-process to
+        the server, so a QA runner — which boots this agent as a subprocess —
+        can never call it and gets `None` no matter how healthy the node is.
+        Logging it here is the only way that value reaches a log tail.
+
+        Why this value and not a derived one: "does this node hold offerable
+        carriers" cannot be answered from the wire (carriers ride the
+        Attestation plane, so there is no Trace envelope kind to count), nor
+        from `trace_events.cohort_scope` (a read-time projection in a different
+        table, downstream of the attestation's own scope), nor from a
+        substring match on the dimension (`trace:` and `trace_summary:` are
+        different namespaces and `covers()` is a prefix test, so a loose
+        `%trace:%` over-counts). persist's `storage_summary()` answers it, and
+        `node_state()` carries that verbatim.
+
+        Purely diagnostic; never disturbs the probe.
+        """
+        try:
+            import json as _json
+
+            import ciris_server  # type: ignore[import-not-found, import-untyped, unused-ignore]
+
+            ns = getattr(ciris_server, "node_state", None)
+            if ns is None:
+                logger.info("[TRACE-PLANE] phase=%s unavailable (ciris_server has no node_state accessor)", phase)
+                return
+            raw = ns()
+            state = _json.loads(raw) if isinstance(raw, str) else raw
+            plane = (state or {}).get("trace_plane")
+            if plane is None:
+                logger.info("[TRACE-PLANE] phase=%s absent (node_state carries no trace_plane)", phase)
+                return
+            logger.info("[TRACE-PLANE] phase=%s %s", phase, _json.dumps(plane, default=str))
+        except Exception as exc:  # noqa: BLE001
+            logger.debug("[TRACE-PLANE] phase=%s accessor error (non-fatal): %s", phase, exc)
+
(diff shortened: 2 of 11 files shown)


emooreatx:
All five addressed in dd7f08ed3. Four of them were defects that would have shipped *looking* fixed.

**The skip still failed the run (P1).** This is the one that mattered most — `_run_sdk_modules` decides success with `"PASS" in status`, so recording the retired endpoint as ⏭️ SKIP left it `success=False`: the module still failed and the QA result stayed nonzero. The endpoint went on breaking the run, it just stopped *looking* like the reason — strictly worse than the red it replaced. Both aggregation sites now share one predicate that treats SKIP as non-failing.

**The fallback could substitute an unrelated peer (P1).** Correct, and it defeats the purpose: with `canonical_targets` non-empty but no target yet in `peers`, `peers[0]` handed back some other federation peer whose `knows_peer`/`kex_present` were then reported as the canonical's — GREEN, under exactly the missing-canonical condition being diagnosed. A named target that is absent now yields `None`.

**`deliverable` was ignored (P1).** Also right, and I should have caught it since the decision tree below already keys on it. Rooting plus KEX doesn't imply deliverable — delivery may not have started, or the edge may be down, with both booleans still true — and setting `kex_state=PRESENT` there printed "trace envelopes can seal" *and* skipped `_diagnose_delivery_status`, the one thing that would have named `deliverable=false`. PRESENT now requires the node not to have said otherwise, and a not-deliverable peer gets its own `kex_state`.

**Untyped dict (P1).** Fair per AGENTS.md, and it matters here specifically: those fields feed a claim about whether traces can flow, and a renamed or absent key reads as `False` — the exact failure mode this PR corrects. Now a `CanonicalPeerState` model, every field `Optional` so unknown stays distinguishable from false, `extra="ignore"`.

**Console printed `{e}` (P2).** Right — I computed `detail`, stored it, logged it, and then printed the empty `str(e)` on the line the operator actually reads. The bug surviving inside its own fix.

Tests for each: unrelated-peer substitution returns None, absent fields stay None, the model ignores extras, PRESENT gated on deliverable, the aggregation predicate, plus source guards that neither aggregation site reverts to a bare `"PASS"` check.

**288 passed** across `tests/tools/qa_runner/`.

emooreatx:
All four live findings fixed in df4682544, plus the emoji that was failing the build and the retry work.

**`carriers` doesn't exist — caught cleanly.** I invented it. `trace_plane` never returns that field (the substring appears nowhere in the 0.5.188 extension, while `standing`, `band`, `last_admitted_at` and `age_seconds` all do), and because the model ignored extras and defaulted to `None`, every real payload validated and the verdict printed `carriers=?` forever — a diagnostic incapable of ever producing a reading. Now modelled and displayed as the four real fields.

**Green needed an affirmative `deliverable`.** Right: `is not False` admitted the unknown case, which is the exact state `CanonicalPeerState` exists to preserve — so a snapshot that merely omitted the field printed "trace envelopes can seal" and skipped the diagnosis on no evidence. Requires `is True` now, with a distinct `kex-yes-deliverable-unknown` state.

**No named target is not an answer.** Also right, and I'd only fixed half of it: I stopped the non-matching-target case but left `peers[0]` for the empty-target case, which is the same manufactured green by another route. Both guessing paths are gone.

**"No peer entry" contradicted the line above it.** An entry that exists but omits `knows_peer` prints a canonical key two lines earlier, so calling it "no peer entry" misdiagnoses an incomplete snapshot as an absent peer. The two unknowns read differently now.

*(The fifth — `node_state()` called in-process from the runner — was already fixed in 7fd7f1680: the node logs `[TRACE-PLANE]`, the runner reads it. That review was against the older commit.)*

**And the emoji.** The `⏭️` in my SKIP marker tripped the repo's own guard — on a Windows cp1252 console it raises `UnicodeEncodeError` and kills the process. `[SKIP]` now, like `[OK]` and `[FAIL]`.

## The retry, which is the point of all this

A live run got exactly **one** reprime: flat 180s cadence, window torn down at 3m16s — 141 rounds, 21 Key envelopes sent, **zero inbound**, KEX never landed, and the node never reached the canonical at all.

- **Front-loaded schedule**: 45s / 120s / 300s, then the 180s steady cadence. Reprime is idempotent (CIRISServer#288), so an early nudge costs one dial and buys the peer a chance to heal its cache while the window is still young.
- **Name the wall**: zero-inbound and inbound-but-no-KEX wear the same `kex_present=false` label, and only the second is worth waiting through. A canonical fail-closes on an unknown `attesting_key_id` and says nothing back (CIRISServer#488), so from the sender's side a structural refusal is indistinguishable from congestion. The probe now reads `envelopes_received_total` and, at zero after two reprimes, says plainly that the peer is not replying and that reprime cannot fix it — the key has to be registered there.

**319 tests pass.**

Issue #1116, opened 2026-08-25
CI: coverage shard merge concatenates instead of unioning — Sonar scores one shard, understating new_coverage


emooreatx:
## Summary

`build.yml`'s "Combine coverage reports" step does not merge coverage — it
concatenates it. Sonar therefore scores every PR against **one arbitrary
shard's** view of each file instead of the union across all 8, so genuinely
covered lines are reported as uncovered and `new_coverage` reads far below
the true number.

## The bug

```python
# Use first file as base
base = ET.parse(files[0])
base_root = base.getroot()

# Merge others
for f in files[1:]:
    tree = ET.parse(f)
    for pkg in tree.findall('.//package'):
        base_root.find('.//packages').append(pkg)
```

Each shard emits a full-tree `coverage-shard-N.xml` covering *every* file
(`--cov=./`), with hits only for what that shard executed. Appending the
`<package>` elements yields a `coverage.xml` containing 8 duplicate entries
per file — one per shard. Nothing ORs the per-line `hits` attributes, so a
consumer resolving the duplicate keys keeps a single shard's counts.

## Evidence (run 32784489152, PR #1114)

`ciris_engine/logic/runtime/edge_runtime.py`, lines 314/315/316 (`should_reprime`)
and 330 (`peer_is_silent`), per-shard:

| shard | covers |
|---|---|
| 1 | 314, 315, 316 |
| 2 | 314, 315 |
| 3 | 314, 315, 330 |
| 4 | 314, 315, 330 |
| 5 | 330 |
| 6 | 314, 315 |
| 7 | none |
| 8 | none |

Union = all four covered. Sonar reported all four **uncovered**, i.e. it
resolved to shard 7 or 8's copy. File-level union is 81/297 = 27.3%; Sonar
scored the file far lower, and the PR's `new_coverage` came out at 14.6%
against an 80% gate.

Locally, `pytest tests/ciris_engine/logic/runtime/test_delivery_retry_policy.py
--cov=ciris_engine.logic.runtime.edge_runtime` confirms 285-345 is covered —
those lines are exercised by committed tests.

## Impact

- `new_coverage` is understated on every PR that touches a file whose tests
  land in a different shard than the alphabetically-first coverage artifact.
- The quality gate fails on work that is actually tested, which trains us to
  admin-merge past Sonar — the failure mode this gate exists to prevent.
- Absolute `coverage` on main is understated for the same reason.

## Fix

[...]

PR #1117, opened 2026-08-26
fix(attestation): a slow verifier bricked the install instead of degrading

emooreatx:
## The report

Francesco's 2.9.37 on **Windows 11 / Python 3.14** died 22 seconds after boot — no UI, no setup wizard, no way in. The whole log spans `17:05:06` → `17:05:28`.

```
17:05:09.555  [tree_verify] ciris_verify_tree(project=ciris-agent,
              root=C:\Users\franc\...\site-packages, version=2.9.37, ...)
              <- never returns
17:05:28.094  ATTESTATION GATE BUDGET BREACH
                exception_message: Startup attestation exceeded the 20s budget (elapsed=20.0s)
                attestation_task_done:  False
                stage_timings_seconds:  {}
17:05:28.698  [EXIT] CIRIS agent exiting cleanly
```

Exactly one `tree_verify` line — it went in and never came out. `stage_timings_seconds: {}` because that dict is only written *after* `run_attestation()` returns.

## The bug: the gate was racing the degrade it was waiting for

`verifier_runner` already promises that *"an attestation is always produced within the budget"* — it bounds its own run at `attestation_deadline_seconds()` and hands back a degraded `level=0, binary=FAIL` result rather than hanging.

The processor gate in `await_attestation_ready` exists to wait for that result. But both were bounded by the **same** budget, measured from within milliseconds of the **same** instant. The gate expired while the runner was still assembling its result — and lost every time. `elapsed=20.0s` is the budget dead on, with the task not yet done.

## Worse: the breach was permanent for the process

`remaining` is measured from task-creation time:

```python
remaining = max(0.0, budget_seconds - elapsed)
```

Once elapsed passed the budget, every later caller got `0.0` and timed out instantly. `batch_context` calls this same method **per thought**, so one slow boot meant the agent could never think again — even after the attestation succeeded a second later.

## The fix

1. **The gate waits `attestation_deadline_seconds() + 5s grace`**, so the runner's bounded degrade can land. This is not more budget for a slow verifier: past deadline+grace the runner's *own* timeout has failed, which is a different bug and is reported as one.

2. **A breach is reported, not raised.** A slow filesystem does not make the tree hash wrong. Warned once, then DEBUG — per-thought callers land here too and would otherwise bury the one that matters.

3. **The processor starts degraded rather than aborting**, with the receipts dump unchanged.

## Why this is not a security regression

The gate is a **latency pre-warm, not the enforcement point**. `batch_context` blocks on the same `await_attestation_ready()` and refuses to build a thought without an attestation result:

```python
if attestation_result is None:
    raise RuntimeError("CRITICAL: No attestation result available ...")
```

So nothing runs unattested either way. Starting degraded costs first-thought latency; aborting cost the whole application. The 20s budget is unchanged and still loud when blown.

## Tests

New `test_attestation_gate_does_not_race_degrade.py` (6 cases) locks:
- the gate deadline strictly outlasts the runner deadline — the defect in one assertion
- a degrade landing just past the budget is still delivered (fails under the old code)
- a breach does not poison later callers
- slow-but-successful is reported, not raised
- the deadline still tracks a late `CIRIS_STARTUP_ATTESTATION_BUDGET_SECONDS` override

`test_attestation_gate.py` updated to the new contract — receipts assertions unchanged, `pytest.raises` wrappers removed.

748 passed across `tests/ciris_engine/logic/runtime/`, `tests/ciris_engine/logic/services/infrastructure/`, and `test_batch_context.py`.

https://claude.ai/code/session_01N1ugR27TCMG2ZWQK6H8knJ

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -252,6 +252,17 @@ jobs:
           # repo. SIGTERM at 45m inside a 60m job leaves room for that; sizing
           # the inner bounds closer to 60 would recreate the exact failure above,
           # where the outer bound wins and the inner ones never report.
+          # COVERAGE_FILE: emit the raw coverage DATA file, not a per-shard XML
+          # report. The old flow wrote 8 full-tree XMLs and "merged" them by
+          # appending <package> elements, which never OR'd the per-line hits, so
+          # Sonar resolved each file to ONE arbitrary shard's view and scored lines
+          # as uncovered that other shards had covered (CIRISAgent#1116).
+          # `coverage combine` unions properly, and is the only thing that can
+          # merge BRANCH data at all.
+          #
+          # No leading dot in the name: upload-artifact skips hidden files unless
+          # include-hidden-files is set, which would silently drop the data.
+          COVERAGE_FILE=coverage-data-shard-${{ matrix.shard }} \
           CIRIS_TEST_CONTROLLER_DEADLINE_SECONDS=2580 \
           timeout --signal=SIGTERM --kill-after=30 2700 \
             pytest tests/ -n 4 --dist=loadgroup \
@@ -260,13 +271,13 @@ jobs:
             --store-durations --durations-path=test_durations_shard_${{ matrix.shard }} \
             --reruns=2 --reruns-delay=1 --max-worker-restart=4 \
             --session-timeout=2520 \
-            --cov=./ --cov-report=xml:coverage-shard-${{ matrix.shard }}.xml
+            --cov=./ --cov-report=
 
       - name: Upload coverage artifact
         uses: actions/upload-artifact@v6
         with:
-          name: coverage-shard-${{ matrix.shard }}
-          path: coverage-shard-${{ matrix.shard }}.xml
+          name: coverage-data-shard-${{ matrix.shard }}
+          path: coverage-data-shard-${{ matrix.shard }}
           retention-days: 1
 
       - name: Upload durations artifact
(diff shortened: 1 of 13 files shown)


emooreatx:
Addressed all three Codex findings in 3aa4cc1ad, and folded in the coverage-merge fix (#1116) since this PR is what surfaced it.

### 1. Propagate the breach to the receipt logger (P2) — valid, fixed

Real regression. The gate stopped raising, so `_await_attestation_ready` logged *"Startup attestation complete"* and never entered the branch that dumps receipts — losing the diagnostics for the exact wedged-verifier case this path exists to serve.

`await_attestation_ready` now returns a typed outcome and the gate drives receipts off it:

```python
class AttestationGateOutcome(str, Enum):
    READY = "ready"
    SLO_BREACH_COMPLETED = "slo_breach_completed"   # valid result, arrived late
    SLO_BREACH_PENDING = "slo_breach_pending"       # runner's own degrade didn't land
```

Three new cases in `test_attestation_gate.py` lock it: a pending breach dumps receipts *and* degrades, a completed breach still files, and a healthy boot files nothing and does not claim success falsely.

### 2. Honour the caller's gate budget (P2) — valid, fixed

Also real, and worse than described: `await_attestation_ready(budget_seconds=0.05)` waited ~25s.

The grace exists for exactly one reason — to outlast `verifier_runner`'s own bounded degrade so its result can land. A caller asking for *less* than the runner's deadline isn't racing anything, so it's honoured exactly:

```python
if budget_seconds < attestation_deadline_seconds():
    return budget_seconds
return budget_seconds + ATTESTATION_GATE_GRACE_SECONDS
```

You were right that this broke `tests/services/infrastructure/test_attestation_refresh.py` — 6 failures. That suite lives under `tests/services/`, not `tests/ciris_engine/`, which is why my sweep missed it; the "748 passed" in the original PR body did not cover it. Those tests encoded the old raise-on-breach contract and now assert the outcome plus the logged warning, keeping each test's original point — including the load-bearing *"names the verifier as the thing to fix, not the budget"* assertion, now checked against the warning text. That suite also dropped 40.7s → 15.8s, which is the budget fix showing up directly.

### 3. Type the stage-timings fixture (P1) — fixed

`Dict[str, float]`.

### Also: CIRISAgent#1116, the coverage merge

Worth calling out because it is why this PR's Sonar number was wrong. The combine step appended `<package>` elements from 8 per-shard XMLs without ever OR-ing per-line hits, so Sonar resolved each file to **one arbitrary shard's view**. On run 32784489152, `edge_runtime.py` 314-316 and 330 were covered by shards 1-6 and scored uncovered.

Shards now emit raw coverage data and the job runs `coverage combine` — which unions line hits and is the only thing that can merge branch data at all. Verified locally that two genuinely different shard files combine to exactly the union, with the lines only one shard covered surviving. A missing shard is now reported instead of silently understating the total, and the data file is deliberately not dot-prefixed (`upload-artifact` skips hidden files unless `include-hidden-files` is set).

`tests/ci/test_coverage_shards_are_unioned.py` pins the contract — the old failure was silent, and a wrong coverage number looks exactly like a right one.

**237 passed** across the attestation and CI suites; **1135 passed** across `logic/runtime/`, `logic/services/infrastructure/`, `tests/services/`, `schemas/`, and `test_batch_context.py`.

PR #1118, opened 2026-08-26
fix(quality): clear the Sonar security + reliability gates

emooreatx:
Main has been failing its quality gate on three conditions: Reliability 3, Security 3, Coverage 49.2%. This clears the two ratings. (Coverage is a separate matter — the 49.2% figure was itself wrong, fixed in #1117 / #1116.)

## Security: 3 → A (16 vulnerabilities)

**Three were real and are fixed.** `name` is an unvalidated FastAPI path param on `PUT /providers/{name}/priority`, f-string-interpolated into four log lines — a crafted provider name could forge log entries (CWE-117). Converted to lazy `%r` formatting, which is the pass this repo already applied across ~10 modules. Same for `adapter_name` in `llm_providers.py`.

**Thirteen were marked with reasoning**, not silently dismissed:

| Count | Site | Why not a vulnerability |
|---|---|---|
| 9 + 1 | `compose_dump.py`, `research_overrides.py` | Developer CLIs under `if __name__ == "__main__"`, imported only by `tools/research/*` and tests. No adapter, route, handler or tool_service reaches them, so no LLM can drive these paths — and the operator already has filesystem access, so there is no privilege boundary to escape. S8707 targets LLM-invoked file operations. |
| 1 | `wizard.py:366` | Writes to `get_default_config_path()` — a computed value, never input. No traversable path to validate. |
| 2 | `authentication_store.py`, `budget_envelope.py` | Already mitigated **at the flagged line** — lazy `%r`, with comments citing CWE-117/S5145. The scanner is flagging the fix. |
| 1 | `authentication/service.py:1636` | Now points at a function signature; the file moved. |

## Reliability: 3 → A (130 production bugs)

The bulk was **122× S7497 — `asyncio.CancelledError` swallowed**. That is a real defect class: a coroutine absorbing its own cancellation completes normally, so shutdown cannot distinguish a torn-down task from a finished one.

**But it is not one defect class, and a blanket fix is wrong.** I verified that the hard way — my first pass applied `raise` everywhere, and adding it to `AccordMetricsAdapter.stop()` meant the next line, `await self.metrics_service.stop()`, never ran.

```python
# WRONG to re-raise here — this code cancelled the child itself
self._lifecycle_task.cancel()
try:
    await self._lifecycle_task
except asyncio.CancelledError:
    pass          # correct: absorbing a cancellation you initiated
```

- **63 sites now re-raise** — these swallow cancellation of their *own* coroutine.
- **61 sites keep absorbing**, marked won't-fix upstream with that reasoning — these call `<task>.cancel()` then await that same task only to confirm teardown.

Also fixed:

- `balance_monitor` poll loop: `break` on cancellation left the task looking completed → re-raises.
- `external_data_sql`: two synchronous `open()` calls inside `async` functions moved off the event loop via `asyncio.to_thread`, matching `profile_loader.py`'s pattern.
- `external_data_sql`: **two byte-identical if/else branches collapsed.** The comment claimed `TimeServiceProtocol` had no suitable method — it has `now_iso()`. The service was injected and then ignored, so a *signed GDPR Article 17 deletion proof* carried a wall clock no test could control.
- `accord_metrics`: float-equality sentinel (`!= 0.0`) → `Optional[float]`.
- `sample_adapter`: a caller's `redirect_uri` was silently overwritten.
- **3× S5779**: mypy-narrowing `assert`s inside `try` blocks that catch `Exception` — the `AssertionError` would be swallowed as an ordinary error, and asserts vanish entirely under `python -O`. Replaced with local bindings that give real narrowing.

`test_run_lifecycle_handles_cancellation` now asserts the corrected contract: *gracefully* means the `finally` cleanup still runs, not that the cancellation disappears.

Test files are excluded from both ratings (`sonar.tests=tests`), so the 25 test-file findings are untouched by design.

## Verification

Targeted suites over every touched area pass: 240 (adapters), 340 (shutdown), 100 (llm routes/providers), 58 (adaptive_filter), 6 (accord_metrics lifecycle).

A full-suite run recorded **zero failures** across ~11.9k tests but died in an xdist `pytest_sessionfinish` teardown (`OSError: cannot send`), losing the summary line — so I am not claiming a clean exit code from it. CI is the authority here.

https://claude.ai/code/session_01N1ugR27TCMG2ZWQK6H8knJ


emooreatx:
## Sonar vulnerability dispositions — and a correction to my own earlier reasoning

The PR's new-code scan surfaced 4 vulnerabilities. All are now dispositioned, PR vulnerabilities are **0**, and both new-code ratings are **A**. Recording the reasoning here because Sonar's API refuses comments on PR-scoped issues (403), so it would otherwise live nowhere.

### The correction, first — I got two markings wrong

In the initial pass I marked two S5145 findings **false positive**, claiming they were "already mitigated at the flagged line" because they use lazy `%r` formatting, and that the scanner was flagging the fix.

**That reasoning does not hold.** S5145 is a *taint* rule: it flags user-controlled data reaching a log sink at all, regardless of escaping. `%r` changes the content written but does not break the taint path.

This PR proved it. I applied the same `%r` treatment to `llm_routes.py` and `llm_providers.py` as a *fix*, and Sonar flagged both again as new-code findings — the same rule, on the code I had just "fixed".

I reopened both mis-marked issues, recorded the correction on each, and re-dispositioned them as **won't fix** on accurate grounds.

### The accurate disposition for S5145 (4 issues)

The attack the rule names *is* prevented: `%r` escapes newlines, CR and non-printables, so a crafted identifier cannot forge a log line (CWE-117). What the rule actually objects to is that a user-controlled value appears in the log **at all**.

Satisfying that means not logging the identifier — and the identifier is the entire diagnostic value of these lines (*which* provider, *which* adapter, *which* `wa_id`). So this is an accepted residual, not a fix: the value is in the log, escaped, and the log is not a trust boundary we cross.

If that trade is wrong, the fix is to drop the identifiers, not to re-escape them.

### S5332 ×2 — false positive, RFC 8252

`sample_adapter/configurable.py:153`, "Using HTTP protocol is insecure."

`LOOPBACK_REDIRECT_HOST = "127.0.0.1"` (`configurable.py:52`), and **RFC 8252 §7.3 specifies `http://127.0.0.1:{port}/`** as the redirect URI for native applications. Loopback traffic never leaves the device, so there is no transport to intercept, and HTTPS would require a certificate for an address no CA will issue. The adjacent comment already recorded this. It surfaced as *new* code only because an adjacent fix moved the line — the loopback URI predates this change.

### Remaining

The gate still fails on `new_coverage: 16.3`. That is now a *true* number (the shard-merge defect that understated it was fixed in #1117/#1116). The uncovered new lines are the 63 `raise` statements in adapter cancellation handlers — paths that execute only on cancellation during shutdown. Not chasing it per direction; flagging it as the one condition still red.

Main-branch vulnerabilities show 3 open: those are the three I fixed in code here, and they clear when main is re-analyzed post-merge.

PR #1121, opened 2026-08-30
2.9.43 — adopt ciris-server 0.5.196 + ciris-client 0.5.195: the node identity works, and the new login UI ships

### LINKED ISSUE #1089: Adopt ciris-client[agent] and delete client/ — the extraction's terminal state
The agent half of CIRISServer#471's flip (see that issue for the full account): \`ciris-client[agent]\` in requirements, \`client/\` deleted, \`CIRISBuild.HAS_AGENT\` taken from the wheel's flavor instead of a hand-flipped const (the \`-PhasAgent\` build input already exists in CIRISClient). The fix for the 141 placeholder-corrupt translations (#1085 / PR #1086) is already folded into CIRISClient's tree via the three-way merge, as are the destructive locale-\`Sync\` removals — adopting the package retires that whole class of hand-carried alignment.

Evidence predicate: \`ciris-client\` in this repo's requirements scores the row in CIRISClient's \`evidence/blocked_upstream.tsv\`.

https://claude.ai/code/session_01H4tXJFQ5qYHSjHeT2ePgzz

### LINKED ISSUE #1119: Adopt ciris-server 0.5.189 — one new call before init_edge_runtime, and your node stops running on the brain's key
# Adopting ciris-server 0.5.189 — the mesh reliability cut

Tagged `v0.5.189` (`ca4896b`). Substrate: **persist v38.6.0 / edge v18.10.0 /
verify v13.6.1**. Client: **ciris-client 0.5.189** (server and client versions
now move together, and a release gate enforces it).

Two of your production findings close in this cut, and one thing changes at your
call site.

---

## 1. THE ONE CODE CHANGE: provision before edge init

One new line, between building the Engine and calling `init_edge_runtime`.

```python
node_key_id = ciris_server.provision_node_identity(
    keystore_alias,      # the same alias you pass Engine(...)
    identity_dir,        # the same identity_dir you pass Engine(...)
    actor_key_id=None,   # your bootstrap key_id if this node carries a brain
)

ciris_edge.init_edge_runtime(
    engine, ...,
    use_node_identity=True,
    node_identity_dir=identity_dir,
)
```

**Order matters and is not incidental.** Edge resolves the node key by
`open_existing` and REFUSES when it is absent — correct, because a key edge
minted would be registered by no directory and owner-bound by nobody
(CIRISAgent#1009's shape). Edge inits before CIRISServer's compose folds on, so
without provisioning a first boot has nothing to open.

No key material crosses the boundary: a `key_id` comes back, the signer stays in
Rust. Same property `resolve_user_signer` and `ConsentGrantOptions::author_signer`
hold.

Idempotent — every boot after the first re-opens and re-registers to the same
row, so call it unconditionally.

### Pass `actor_key_id` if the node carries a brain

It makes provisioning verify your actor key is registered and IS an actor. Skip
it and you lose that check, which is the one that catches the 3,404-refusal state
(`attesting_key_id … does not exist in federation_keys`) **at startup, in front
of whoever is starting the node**, rather than in a peer's log they cannot read.

### What provisioning gates, and what it deliberately does not

| requirement | gated | why |
|---|---|---|
| node key registered `node`-only | **yes** | it is the key that walks the lightnet door |
| actor key registered as an actor (when `actor_key_id` given) | **yes** | an unregistered attester has every row refused |
| owner-binding present | **no** | a fresh node has no owner until someone completes OAuth and claims it |

Ownership is excluded on purpose. Requiring it would mean a node cannot start its
transport until claimed. Verified safe: `oauth.rs` has zero edge references, the
claim path touches edge exactly once through a fire-and-forget
`pull_owner_testimony` whose own comment says the response "must not wait on a
peer being reachable", and LLM checks are outbound HTTPS.

---

## 2. WHY: your node has been running on the brain's key

Observed on a production node:

```
key_id         ciris-agent-bootstrap-mplbdbzbed
identity_type  agent
self_key_id=ciris-agent-bootstrap-mplbdbzbed
Reticulum transport identity: … key_id=ciris-agent-bootstrap-mplbdbzbed
```

One key answering two questions. The Reticulum transport identity, `self_key_id`
and the de-admission self all named a key carrying **agency**.

`compose::register_self_key` had tried to register that id as `node` and could
not: persist refuses with a typed `Conflict`, and CIRISServer mapped it to
`Ok(())` at debug under a comment calling it benign. Benign is right for
re-registering an identical trust-root row; it is exactly wrong when the differing
row says `agent`. **persist told us, at debug, and we had already decided it did
not matter.**

**Not fixable by adding `node` to that key's role set.** persist's agency gate
constrains a recipient resolving to a **node-only** identity, so a `{node,agent}`
hybrid PASSES it. Fusing the roles does not blur "infrastructure must not have
agency" — it switches the rule off. That is why CC 3.4.7.3 Clause A makes `node`
non-cohabitable with `agent`/`user`, and why the cure is a second KEY.

Your bootstrap key is **never modified**. It keeps its type, its signatures and
every row it authored. Its envelope binds `identity_type` inside the signed bytes
(CIRISPersist#659), so re-typing it would mean re-minting through an accord scrub
AND would make every historical trace retroactively node-authored.

---

## 3. Your production findings, closed

**The duplicate storm** (CIRISPersist#771). 7,536 attestation envelopes refused in
six hours on canonical — 428 distinct rows re-sent up to **82× each** — because
`put_attestation` reported an already-held row as `federation_backend`, a backend
ERROR. The sender never learned the row had landed. 58% of all refusals on that
node. persist v38.6.0 returns `AlreadyHeld`; edge routes it to
`ApplyOutcome::Duplicate` — counted, not swallowed.

**Unregistered attesting keys** (the other 5,443). `datum-j7xp3i7quf` (3,404),
`echo-speculative-4fc6ru-*`, `echo-core-jm2jy2-*`. Not a substrate bug — those
keys were never admitted to `federation_keys`. Passing `actor_key_id` to
provisioning now catches that class at startup.

**CC 3.4.7.3 graduation** (CIRISPersist#773). The agency gate is a membership
test, `may_act_through(agent, node)` exists, 28 citations re-pointed.

---

## 4. The tool-envelope seam (CIRISServer#492)

Your DX request stands and this cut does not close it. Correcting one fact from
our reply, since it changes the size of the ask:

**`authorize_delegated` EXISTS** — `src/auth/gate.rs:162`, implemented, with
`DenyReason { denied_by, verb, goal, detail }`, check order `never → deny →
allow`, and a `never_delegatable` floor (`Delegate | Wipe | AccordHalt |
ChatAuthor`). Your `hasattr` check returned False because it is a **Rust function
not exported through PyO3**, not an absent primitive.

But your FSD is right that it is barely wired: **one caller**
(`contacts_chat.rs:289`). So the ask is "add verbs and call sites", not "build an
authorization engine".

Also note **`may_act_through` is now built** (persist v38.6.0), and it is
fail-closed in both directions the ruling demanded: `AmbiguousNodeOwner`
propagates rather than collapsing to `false`, and a node with no owner-binding
returns `Err`, not `Ok(false)`. Your point 6 — "denials that cannot reach a human
should be denials" — has its predicate now.

---

## 5. Two things on your side, unchanged by this cut

- **`compliance/D19_partner_role.md:66` and `compliance/README.md:131-132`** still
  say `requires_approval` "gates nothing". False at your HEAD —
  `thought_processor/main.py:257` branches on it. Prose conceding a closed gap is
  its own hazard: someone will design around a hole that is not there. We hit the
  identical shape the same week (`field_conformance.rs:349` still says
  `authorize_delegated` is "PROPOSED, not yet wired"), so this is a class, not
  two incidents.
- **`LESSONS_LEARNED.md`** still recommends Haiku sub-agents for translation —
  the guidance that produced the word-salad the client team had to undo.

---

## 6. Release notes

`FSD/ACTOR_NODE_KEY_SPLIT.md` in CIRISServer carries the full mechanism: which key
signs what, the lightnet/darknet mapping, the CIRISServer#312 constraint, and the
migration order.

Related: CIRISConstitution#95 (CC 3.4.7.3) · CIRISPersist#771 / #773 ·
CIRISEdge#541 · CIRISServer#489 / #492

emooreatx:
## What this adopts

```
ciris-server==0.5.196    the node-identity line, incl. the alias parameter
ciris-client==0.5.195    CIRISClient#23 — federation sign-in removed
```

No `ciris-client` 0.5.196 is published; 0.5.195 satisfies the server's
`ciris-client>=0.5.190,<0.6` and carries the login-screen fix.

**2.9.42 shipped client 0.5.194 and therefore did not contain the new login UI.** I held that pin saying 0.5.195 had no GitHub release; by the time I merged, it had been published for twelve hours. This is that correction, plus the server line that came after it.

## The node identity finally works

0.5.195 broke folded-agent boot: `provision_node_identity` writes
`ciris-node-bootstrap-<fp>` while edge, given only a directory, opens
`<engine alias>-node`. Neither exists for the other, so edge refused
(correctly — CIRISEdge#541) and initialization died.

0.5.196 added **`node_keystore_alias`** so the caller can settle the name. We now pass it, taken from what provisioning already returns (`<node alias>-<fingerprint>`), so the name comes from the side that knows it instead of two derivations hoping to agree. Feature-detected off the signature, with a loud warning on substrates that lack it.

Verified on a clean boot, not inferred:

```
[NODE-KEY] node identity provisioned: ciris-node-bootstrap-yd4n36q6me
[NODE-KEY] passing node_keystore_alias='ciris-node-bootstrap'
Edge runtime initialized: actor=ciris-agent-bootstrap-bam2oudndl
                          node=ciris-node-bootstrap-yd4n36q6me
[OK] Initialize Edge Runtime initialized
Node fold: substrate read-API LISTENING on 4243 [OK]
GET /v1/identity      → identity_type=node
GET /v1/system/health → role=agent, agent={"folded":true,"reachable":true}
```

Two things that were previously only asserted are now demonstrated:

* **actor and node are distinct keys** — CIRISAgent#1119's whole point. A bare node no longer carries the agency-bearing key.
* **`/v1/identity` states its kind** — CIRISServer#507. Nothing has to infer identity kind from a keystore alias that reads "agent".

## Also here

* Two localization guards were scanning `client/shared/src`, deleted in the migration. `rglob` on a missing directory yields nothing, so both passed over zero keys. They now scan the Kotlin this repo still owns and skip loudly when there is none.
* `[NODE-KEY] identity name mismatch` diagnostic — when the two derivations disagree, one line names both, instead of two facts four hundred log lines apart.

## Expected reds

CodeQL (vendored third-party under `apps/ios/Resources/app_packages/**`) and Coverage on New Code — the same two accepted on #1120.

https://claude.ai/code/session_01PzZQ4uvxyTcz1grsvSr3vn

---

## Closes

Closes #1089 — the extraction's terminal state. `client/` is gone (4,363 files),
`ciris-client==0.5.195` is the pin, and `CIRISBuild.HAS_AGENT` no longer exists
in this repo at all — it comes from the wheel, so there is no hand-flipped const
left to drift. One deviation worth stating rather than glossing: the issue asks
for `ciris-client[agent]`, and **that extra does not exist**. `ciris-client`
0.5.195 publishes `web` and `readiness` only; the agent flavor is carried by the
gradle `-PhasAgent` input and `tools/dev/vendor_desktop_jar.py` instead. The
substance is adopted; the literal extra was never shipped upstream.

Closes #1119 — actor and node are distinct keys, demonstrated on a real boot:
`actor=ciris-agent-bootstrap-…` / `node=ciris-node-bootstrap-…`, with
`GET /v1/identity` reporting `identity_type=node`. The provision-before-init
call the issue asks for is in place, and 0.5.196's `node_keystore_alias` is
passed so the node stops running on the brain's key.


emooreatx:
All four addressed in 49dd940a9 — thank you, the P1 and the localization one were both genuinely bad.

**P1 · uncommitted iOS edits.** Correct, and worse than a revert: `git checkout --` restores the committed copy over *whatever is there*, so local work under `apps/ios/Resources` (2,512 tracked files) was discarded with no prompt and no undo. Pointing `IOS_APP_DIR` at the live app is what armed it.

Rather than refusing a dirty tree or saving/restoring, extraction now **skips every tracked member** — the archive supplies only what git does not track, which is exactly the build output (`app_packages/`, `python/`) materialisation exists for. Nothing tracked is written, so there is nothing to restore and no dirty-tree case to refuse; it works in a dirty tree instead of blocking on one. A tracked file *missing* from disk is reported rather than replaced from the stale archive, since a deletion is working-tree state too.

Pinned by `tests/tools/test_ios_resources_materialisation.py`, checked against the old implementation first — `test_uncommitted_edits_survive` fails there with the caller's work gone.

**P2 · transactional wheels.** Correct on both halves. Every ABI now downloads to a temp dir and is size-checked before the live directory is touched; on failure the existing set is untouched and the status propagates as `FAILED`. The old warning-and-SUCCESS only moved the discovery to `check_version_alignment.py` anyway, since the gradle pin is deliberately not bumped.

**P2 · post-refresh commands.** Fixed to `cd apps && ./gradlew :android:assembleDebug` and `apps/android/build/outputs/apk/debug/android-debug.apk`, both verified against `build.yml:1800-1805`.

**P2 · vacuous localization guard.** You're right, and this one stings — the skip I added to *prevent* a vacuous pass recreated it one level up. `apps/android/src` and `apps/ios` both still exist (platform shells), so `if not roots` never fired, while the eight Kotlin files under them contain zero call sites. Measured on this checkout: 8 files, 0 matches, `{}` returned, five tests green on an empty set. The skip is now keyed on whether a reference was actually **found** — a directory existing proves nothing about what is in it. Those five now report `SKIPPED` with the file/match counts in the reason.

emooreatx:
Round two addressed in `43642bf11` + `515ef27b2`. Two of these four were defects in the fixes from the previous round, which is worth saying plainly.

**APK path after `cd apps`** — mine, one commit old. I fixed the stale `cd client` and left a repo-root APK path behind a `cd` that persists in the caller's shell, so step 3 resolved as `apps/apps/android/…`. Gradle now runs in a subshell — `(cd apps && ./gradlew :android:assembleDebug)` — so all three lines paste independently from the repo root rather than depending on each other's cwd.

**Every required ABI before pruning** — also mine, and the more interesting miss. Staging closed the *transient failure* hole and I stopped there; you're right that an incomplete upstream release walks straight through `if not android`. That path is worse than a network blip, because it produces an APK that builds cleanly, pins successfully, reports success, and then fails at import only on the devices whose ABI is missing. Now every entry in `lib.abis` must be published before `WHEELS_DIR` is touched (normalising `arm64-v8a` against the wheel tag's `arm64_v8a`).

**Binding failures on both platforms** — you flagged iOS; Android had the identical shape two functions up, so both are fixed. The iOS case is the one with teeth: the native module is already copied by then, so discarding that `False` let the run rebuild `Resources.zip` and write `substrate.lock.json`, stamping the bundle *current* for a version whose Python wrapper is older than its native API. The lock is what answers "which substrate is iOS shipping?", so a confidently wrong one is worse than none. Both now return `FAILED`, which is what keeps the lib out of `ios_ok`/`android_ok` and gates the rebuild, the lock write, and the verify.

**BUILD_INFO** — confirmed stale exactly as described: `b590002ff9d4` recorded against `a7be77887db1` live. Regenerated last, after every Python change; it now records `43642bf11`.

One thing your finding surfaced that's worth recording: `tools/dev/grace_precommit.py` is supposed to refresh BUILD_INFO during commit and **did not fire on either preceding commit**, so this wasn't a one-off oversight on my part — the automation that should prevent it isn't running. I've noted it in the commit message rather than chasing it here, since it's pre-commit plumbing rather than a release blocker.

New tests: `tests/tools/test_android_wheel_staging.py` covers all four wheel paths. I wrote the incomplete-release case twice — the first mock returned the metadata JSON as the download payload, so the *size* check rejected it and the test passed with the ABI guard deleted. It now returns correctly-sized payloads so the guard is the only thing that can fail it, verified by deleting the guard and watching that one test go red. Same negative control on the P1 materialisation test last round.

emooreatx:
Round three addressed in `b46791bff`. Both findings are in my own fixes from rounds one and two — worth saying, because the pattern is that each fix I made had a hole one layer down.

**Fail closed on `git ls-files`.** This is the P1 coming back through a failure mode, and you're right that it's the same bug. The whole protection rests on knowing which paths git tracks — those are the ones never extracted over. I never checked the query's status, and an unchecked `subprocess.run` answers *"nothing is tracked"* for every way it can fail: dubious-ownership rejection, missing git, corrupt index. Empty then means every tracked file is extractable again, and the stale archive copies land on uncommitted work — in exactly the unusual checkout where nobody would expect it.

An empty result is only trustworthy when git said so *successfully*. A nonzero return now aborts before anything is unpacked, and names the likely remedy (`safe.directory`) since dubious-ownership is the probable trigger.

**Aborting the aggregate iOS rebuild.** Also correct, and the more subtle of the two. Returning `FAILED` per library keeps it out of `ios_ok`, which *reads* like enough — but `Resources.zip` is one tree every library writes into, and the failed leg has already copied its new native module in. Rebuilding because a different library succeeded commits a mixed bundle over a committed artifact and then writes `substrate.lock.json` asserting the successful library is current inside it. The nonzero exit afterwards doesn't undo that.

iOS finalisation is now skipped entirely when any iOS leg failed. I pulled the rule out as `ios_legs_that_failed()` so it's testable without driving argparse, and because it's subtle enough to deserve a name. Android failures deliberately don't block iOS — nothing Android writes goes near that tree — and `PENDING` isn't failure, since nothing was written at all.

Two things stated plainly rather than glossed:

- `test_ios_finalisation_gate.py` tests the **rule**, not that `main()` honours it. The call site is one line and visible at `update_substrate_libs.py:1812`, but that wiring is not under test.
- The fail-closed test *was* checked against the unchecked version, where it fails with the caller's work overwritten. I've run that negative control on every test this review produced, after the first version of the ABI test turned out to pass with its guard deleted.

BUILD_INFO regenerated last again (`6fc16028a58d`).

PR #1124, opened 2026-08-31
fix(ci): fetch the CIRISClient AAR before building the APK — v2.9.42 published none

emooreatx:
## The Android release job cannot build, and has not since the client migration

`v2.9.42` published **no Android APK**. `v2.9.43` is failing the same way right now.

```
Execution failed for task ':android:checkDebugAarMetadata'.
> Could not find :ciris-client-0.5.194:.
  Searched in: apps/android/libs/ciris-client-0.5.194.aar
```

`apps/android/libs/*.aar` is gitignored on purpose — a ~15 MB pre-built binary of another repo's source, exactly what the repo-size rule excludes (*"distribute via GitHub Releases and fetch on install"*). So a clean CI checkout has no AAR and Gradle cannot resolve it.

## The fix

`tools/fetch_client_artifacts.py` was written for this and **was never wired into any workflow** — nothing under `.github/` referenced it. It resolves the version from requirements.txt's `ciris-client==` pin, so the AAR cannot drift from the client inside the wheel.

It shells out to `gh release view`, which needs a token on a runner even for a public repo, so the step passes `GITHUB_TOKEN`. Without it the fetch fails auth and the build fails one step later anyway — different error, same missing APK.

Verified the fetcher works from a clean state by moving the local AAR aside: it re-downloaded the identical 16,000,525-byte artifact.

## Why it went unnoticed, and why the guard is a test

The APK is built **only in the release job**, so no pull-request check exercises it. A missing artifact looks identical to a release that never had one — the next person to notice is whoever goes looking for the APK.

`tests/workflows/test_android_release_fetches_client.py` asserts the fetch runs *before* `assembleDebug` and carries a token. Both assertions were checked against the pre-fix workflow, where they fail.

One of the two initially passed on the pre-fix workflow **for the wrong reason**: `_step_index` returns `-1` when the step is absent and `steps[-1]` is the *last* step, so it asserted against an unrelated step's env. Fixed to require a real index.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1793,6 +1793,31 @@ jobs:
           echo "version=$VERSION" >> $GITHUB_OUTPUT
           echo "Version: $VERSION"
 
+      - name: Fetch the published CIRISClient AAR
+        # WITHOUT THIS THE APK CANNOT BUILD, and it has not since the client
+        # migration. `apps/android/libs/*.aar` is gitignored on purpose — it is a
+        # 15 MB pre-built binary of another repo's source, exactly what the
+        # repo-size rule excludes ("distribute via GitHub Releases and fetch on
+        # install"). So a clean CI checkout has no AAR at all, and Gradle fails
+        # to resolve it:
+        #
+        #     Execution failed for task ':android:checkDebugAarMetadata'.
+        #     > Could not find :ciris-client-0.5.194:.
+        #       Searched in: apps/android/libs/ciris-client-0.5.194.aar
+        #
+        # That is why v2.9.42 published no Android APK. The failure is in the
+        # release job only, so nothing on a PR catches it — the artifact simply
+        # never appears, which looks the same as a release that had none.
+        #
+        # The version comes from requirements.txt's `ciris-client==` pin, so this
+        # cannot drift from the wheel's client.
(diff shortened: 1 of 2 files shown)


emooreatx:
Addressed in `0e39d4691` — you're right, and the rule you cite is the repo's own.

`WorkflowStep` / `WorkflowJob` / `Workflow` are TypedDicts covering only the keys this guard reads, `total=False` where the Actions schema genuinely makes them optional (a `uses:` step has no `run:`, most steps have no `env:`). It earns its keep here rather than being ceremony: `yaml.safe_load` returns `Any`, so without it a typo like `step["evn"]` is invisible to mypy — in a file whose entire purpose is catching a mistake nothing else catches.

One deviation from the obvious implementation, worth stating: my first pass imported `TypedDict` from `typing_extensions`, which is **not a declared dependency of this repo**. The floor is `python_requires>=3.10` (`setup.py:225`), where `typing.TypedDict` is fully available, and 3 of the 4 existing `TypedDict` imports here already come from `typing`. Switched before pushing — leaning on an undeclared transitive dep would have been a latent break in a minimal environment.

mypy clean on the file, and the negative control still holds: both assertions fail against the pre-fix workflow, which is the property that makes this guard worth having.

PR #1126, opened 2026-08-31
Five-platform live QA: one flow, five targets, screenshots and a real trace gate

emooreatx:
## What this is

The one CI job that uses the product: install the built artifact, configure a **real** model with a **real** key, drive the **real** UI to the Interact screen, send a message, and require an answer on screen — on all five platforms we ship. Then assert the traces actually reached canonical.

Every other job checks a *part*. In one release this repo shipped an APK that could not build, a client with no XCFramework, and two PyPI wheels short of a set — all behind green ticks, because nothing exercised the path end to end.

## Shape

**Three runner images, five platforms.** `ubuntu-latest` = Linux desktop + Android emulator; `macos-14` = macOS desktop + iOS simulator; `windows-latest` = Windows.

**One flow, five times.** There is a single CIRISClient Compose app exposing one `TestAutomationServer` on 9091, so the automation is written once and selected with `--platform`. Nothing in the workflow contains platform-specific *test* logic — only platform-specific *bring-up*, which is genuinely different (an emulator is not a window).

Two seams make that hold, and leave room for what is coming:
- `platforms.py` — transport + capture, the only place anything asks what it is running on. `capture()` takes a **kind** because both mobile targets already record video natively, so a p2p-chat or video scenario needs no second mechanism.
- `scenarios.py` — named flows in a registry. `interact` is the first; `p2p_chat` and `video` are a new file plus one `register()` line. `ScenarioResult` is PASS/FAIL/**SKIP**, because a bool cannot tell "ran and was fine" from "never ran".

## The trace gate

Keyed to **`replication_envelopes_served_total`**, not `envelopes_sent_total`. The latter counts edge's application/durable send path and is structurally blind to the anti-entropy replication plane that actually carries `trace:*` rows — so a run that lands 15 trace_events reports zero. Keying to it *inverts* the gate. That is CIRISEdge#434; the plane-correct counter is CIRISEdge#433.

`_log_trace_ship()` surfaces it agent-side, because the QA runner boots the agent as a subprocess and can never call an in-process accessor.

## iOS

Builds with **`apps/ios/scripts/rebuild_and_deploy.sh`** — the script that ships to the App Store — not a hand-rolled `xcodebuild`. That script runs `xcodegen`, rebuilds Resources.zip, and lays down the arch-specific simulator Python bundle; skipping any of them produces an app that launches to nothing rather than a build that fails.

## Screenshots

All five capture the successful Interact screen, and the `gallery` job inlines them into one self-contained `index.html` (GitHub cannot render artifact images in a job summary). Missing platforms render as **visible gaps** — a gallery with four tiles must not read as five.

## Status: unproven

This has never run on real hardware. The `pull_request` trigger is **temporary** and must be removed before merge — it boots emulators and spends live LLM tokens. I expect the first runs to fail; that is what this PR is for.


emooreatx:
Addressed in `1ecbd24aa` + `1a66a02d0`. Several of these invalidate results I had already reported as passing, which is worth stating up front rather than burying.

**Mobile was never brought up (P1).** You're right, and it's the worst of these. `desktop-setup --launch` called `run_desktop_first_run_up()` unconditionally, so `--platform android` booted no emulator and installed no APK — it started the **desktop app on the Android port** and I reported the run as an Android pass. Run 3's "Android: full flow green" was the desktop app. A gate claiming five platforms while exercising three is worse than one that admits to three. Bring-up now dispatches on `--platform`.

**The nightly never used the live key (P1).** Also right, and it explains something I had misdiagnosed. Every run so far printed `LLM: mock (no secret available)` and I attributed it to a missing repo secret. It was the wiring: on `schedule` there is no dispatch input, `inputs.use_live_key` is empty, and Actions' loose equality treats empty as false. Only a real `workflow_dispatch` can turn it off now.

**The trace gate failed healthy runs, twice over (P1 ×2).** Both correct. It took the *first* `replication_envelopes_served_total` match, and that counter is emitted repeatedly while delivery is in flight — so a normal run logs `0` early, a positive value later, and the gate read the zero. Any positive sample now counts. And it scanned once, immediately after chat, against a probe that can spend minutes rooting; `--wait-secs 240` polls to a bounded deadline. Verified on your exact scenario (early 0 → later 15): passes.

**A green tile on a red job (P2).** Fixed — the verdict is written after *both* assertions and records `interact` and `traces` separately, so the gallery can't say PASS while the trace half failed.

**Unbounded capture (P2).** Bounded at 60s, with expiry surfaced through the existing non-fatal path.

Three findings landed on code written hours earlier, all mine:

**The adoption path dropped the admin password (P1).** My early return skipped `_update_system_admin_password()`, so a request carrying `system_admin_password` got a response saying both credentials were configured while the default was unchanged. Declining to mint a duplicate is not a licence to drop the rest of the contract.

**The OAuth gate passes vacuously (P1).** This one stings: `assert_one_holder_per_identity.py` was written *to catch* the lockout, and the workflow that runs it authenticates as local `qaadmin`, so it sees zero provider identities and hits an unconditional PASS. The check written to prevent a vacuous pass was vacuous in the same commit. Zero identities now reports **NOT COVERED** with the reason, and `--require-identity` makes absence fatal for a suite that genuinely completes a callback. It still fails the real broken store.

**"Sending message…" would have counted as a reply (P1).** Worse than the string you named: my hand-written blocklist would have accepted `agent.still_processing` ("Still processing. Check back later. Agent response is not guaranteed.") and `agent.error_generic` — the two strings the UI renders *instead of* an answer. A run where the agent errored would have gone green, which is the exact bug class that assertion was rewritten to catch, one layer down. A blocklist of chrome leaks by construction, so the localization bundle is now the blocklist: 3,327 product strings excluded as data. error_generic FAILS, still_processing FAILS, a real reply PASSES.

**Not yet done:** *Exercise the installed wheel instead of the checkout*. You're right that `ServerManager.start()` runs `python main.py` with `cwd=project_root`, putting the repo ahead of site-packages — so the "artifact gate" is testing the source tree. That is a real hole in the gate's premise and needs the runner invoked from outside the checkout; I have not fixed it yet and am not claiming otherwise.

Issue #1127, opened 2026-08-31
Agent SPEAKs successfully but the UI never renders the reply — reproduced on Linux, Android and Windows


emooreatx:
Found by the new five-platform live QA gate (PR #1126), run `33404942070`. Reproduced on **three platforms** in the same run.

## Symptom

Send a message from the Interact screen. The agent processes it, SPEAKs, and the audit records success — **and nothing appears on screen**. The user's own message sits there with its pending indicator.

## Evidence (linux, mock LLM)

```
14:57:24.335  Dispatching action speak for thought th_seed_212ac074_… to handler SpeakHandler
14:57:24.363  [AUDIT] Creating entry for speak:
14:57:24.363  [AUDIT]   outcome: success
14:57:24.363  [AUDIT]   content: [MOCK LLM] Response to user message
14:57:27.845  Updated thought … status to completed
14:57:27.865  Marked parent task … as COMPLETED due to TASK_COMPLETE
```

Sub-second end to end. The UI then polled for **45 seconds** and rendered nothing new but our own echo.

Channel is internally consistent — the task was "Respond to message from @wa-root-qa-node-1788188218-33mixbzmxi" and the reply went to `api_wa-root-qa-node-1788188218-33mixbzmxi`, the same channel. So this is not a channel mismatch.

## Screenshot

The captured Interact screen shows exactly one message — ours — with a pending "…" indicator and `Showing last 1 messages`. Attached to the `live-qa-*` artifacts on the run.

## Why this was invisible until now

`wait_for_response` baselined an ELEMENT COUNT taken before clicking send. Sending adds the user's own bubble to the composed tree, so "the tree grew" went true the moment the echo rendered — 1005ms, which is how long the echo took. The assertion could not tell "the agent replied" from "my own message appeared", and reported `[OK]` over this bug.

That assertion has now been vacuous twice: it began as `await asyncio.sleep(5)` ("we don't have a way to detect response yet"), which was replaced by tree-growth, which merely moved the hole. It now requires new TEXT that is neither our echo nor UI chrome (fixed in PR #1126).

## Scope

- Linux desktop, Windows desktop, Android emulator — all three fail identically
- macOS/iOS not yet confirmed (those jobs were cancelled by concurrency in earlier runs)
- Mock LLM. The live-key path did not exercise here (`LLM: mock (no secret available)`), so this is **not** an LLM or latency problem — the reply existed and was dispatched successfully.

## Suspected area

[...]

PR #1129, opened 2026-09-01
fix(qa): 2.9.45 — five-platform gate green end to end (iOS builds on CI, chat authenticates, trace rung reports)

emooreatx:
Follow-up to 2.9.44, closing the loose ends in the five-platform live gate so it can run green with real screenshots.

## What this fixes

**1. The gate's own 401.** Moving the reply assertion to `/v1/agent/history` gave `desktop-chat` a dependency the other two steps already had — credentials. It had none, defaulted to `admin`, and this workflow only creates `qaadmin`. Every platform went red for a reason that said nothing about the product.

**2. iOS could never have built on CI.** Fixing the hardcoded `/Users/macmini` path got the build to the right directory and into the real obstacle: `ios/` is **untracked** (`git ls-files ios/` → 0 files), so `briefcase build iOS` cannot run on any runner. It doesn't need to — the Python bundle is already fetched by `fetch_client_artifacts.py` (the script's own preflight prints `✓ Resources: 46MB` two lines before failing). `--quick` overlays the repo's live sources onto it, which is the part CI exists to test.

**3. Trace rung reports instead of blocking.** `replication` sits on the KEX/replication path upstream is rebuilding. Downgraded deliberately: still runs, still requires `replication` (never `ship`), emits `::warning::` saying enforcement is off, and the interact half stays fatal.

## The RCA that got here

Three earlier theories, each wrong, each disproved by the next run — recorded because two of them looked convincing:

| theory | verdict |
|---|---|
| client fails to render replies (#1127) | **wrong** — screenshot shows the reply on screen |
| 45s deadline too short | real, but not the cause |
| model arbitrarily chosen (`aion-labs/aion-2.0`, first of 420) | real, and fixed |
| `/tree` is keyed on testTag; transcript untagged | **the actual cause** |

The screenshot that settled it shows the exchange completing normally while the automation reported "nothing new at all — the echo did not render either". Our own sent message was invisible too, which no agent-side bug produces.

CIRISAgent#1127 closed as not-planned (its title asserted the opposite of what happened). Untagged transcript filed as CIRISAI/CIRISClient#27.

## Verification

Five-platform gate dispatched against this branch; screenshots attached to the run's gallery artifact.

https://claude.ai/code/session_01MkVUHTQh2CAQXQpU18SYzV

### DIFF
--- a/.github/workflows/five-platform-live-qa.yml
+++ b/.github/workflows/five-platform-live-qa.yml
@@ -366,11 +366,33 @@ jobs:
           # missing from the artifact. Searching every plausible home and every
           # phase, because the one log that matters is the one for the phase that
           # failed.
-          for d in "${TMPDIR:-/tmp}" "${RUNNER_TEMP}" "$HOME" "${LOCALAPPDATA:-/nonexistent}" .; do
+          # WINDOWS KEEPS IT SOMEWHERE UNIX PATHS DO NOT REACH. The desktop client
+          # writes to `C:\Users\RUNNER~1\AppData\Local\Temp\ciris_desktop_setup.log`
+          # — observed verbatim in the Windows job. `$TMPDIR` is unset there and
+          # `$LOCALAPPDATA` arrives as a BACKSLASH path that `find` cannot walk, so
+          # a Unix-shaped root list quietly collected nothing on the one platform
+          # whose client log we have actually needed.
+          #
+          # cygpath converts when present (git-bash on the Windows runner); the
+          # ${VAR//\\//} fallback covers a bash without it. `$TEMP`/`$TMP` are the
+          # Windows spellings of the same directory.
+          roots="${TMPDIR:-/tmp} ${RUNNER_TEMP} $HOME ."
+          for w in "${LOCALAPPDATA:-}" "${TEMP:-}" "${TMP:-}"; do
+            [ -n "$w" ] || continue
+            if command -v cygpath >/dev/null 2>&1; then
+              w=$(cygpath -u "$w" 2>/dev/null || echo "$w")
+            else
+              w="${w//\\//}"
+            fi
+            roots="$roots $w"
+          done
+          for d in $roots; do
             [ -d "$d" ] || continue
-            find "$d" -maxdepth 2 -name 'ciris_desktop*.log' -newermt '-2 hours' \
+            # depth 3: AppData/Local/Temp sits three below $HOME on Windows.
+            find "$d" -maxdepth 3 -name 'ciris_desktop*.log' -newermt '-3 hours' \
               -exec cp {} artifacts/ \; 2>/dev/null
           done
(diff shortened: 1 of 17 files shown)


emooreatx:
## Codex review — all five addressed (`a7eceb2e5`)

Right on every count. Three were holes I opened, and two of those made the gate report success at the moment the product failed.

| # | finding | resolution |
|---|---|---|
| P1 | Require a new real agent response from history | **fixed** |
| P1 | Restore Python bundle preparation before the iOS build | **fixed** (differently — see below) |
| P1 | Tear down macOS services before exercising iOS | **fixed** |
| P2 | Represent the unobservable trace result separately | **fixed** |
| P1 | Keep the response assertion on the rendered UI | **acknowledged, not fixed — blocked upstream** |

**`is_agent` accepted error text as an answer.** Confirmed in `routes/agent.py`: `if message_type in ("system", "error"): is_agent = True`, set deliberately so the agent does not re-observe its own notifications. So the gate went green on the error row emitted when processing *failed*. Now keyed on `message_type == "agent"` plus a pre-send history baseline, extracted to `_is_new_agent_reply` with a test per failure mode — including one that pins the `is_agent` distinction itself, so a later simplification back to it fails loudly rather than silently reopening the hole.

**iOS bundle** — your diagnosis was right and my `--quick` change would have built an app that launches to nothing. My "✓ Resources: 46MB" evidence was 2,512 *other* tracked files. But `prepare_python_bundle.sh` can't be restored on CI: it copies from a BeeWare app, and `ios/` is untracked (`git ls-files ios/` → 0), so no runner can produce it. Both halves are fetchable in-runner instead: the stdlib ships inside the **tracked** 28M `apps/ios/Resources.zip` (1,210 `python/` entries — `Resources/python` only looks missing because the *extracted* form is gitignored, and `build.yml` already extracts it this way), and `Python.xcframework` comes from `beeware/Python-Apple-support` pinned at `3.10-b14` to match the bundled python3.10 — the same artifact briefcase downloads. Both asserted afterwards, since their absence doesn't fail a build, it produces the silent-launch app.

**Teardown** — ports 8080/9091/8091 cleared between platforms, by port rather than by name since the three processes agree on nothing else.

**Unobservable traces** — exit 3 now, distinct from 0 and 1. It had been returning 0, so `chat-*.json` recorded `"traces": true` and the gallery printed "traces reached canonical" for a run whose own output said NOT COVERED. All three renderings verified against fixtures.

**The UI assertion — you're right, and I can't close it here.** Moving to `/v1/agent/history` does turn a UI gate into a backend check: a client that drops the reply now passes. There is no client-side signal available — `/tree` is keyed on `testTag` and the transcript is untagged, so not even our own echo is visible to automation. Filed as CIRISAI/CIRISClient#27. Until it lands, the screenshot is the only client-side evidence and nothing automated reads it; the failure message now says so rather than implying screen coverage it doesn't have.

PR #1132, opened 2026-09-02
chore(deps): adopt ciris-client 0.5.199

emooreatx:
Bumps `ciris-client` 0.5.196 → **0.5.199**, and carries what adopting it through the five-platform gate turned up. Every finding below has a run, a log line, and an owner.

## The pin (`fb92cc046`)
Held at .197 for CIRISClient#30; .198 fixed it structurally, .199 shipped the mobile half of #28. Artifacts verified against the downloaded assets: AAR (`ai/ciris/mobile` 5880 + `ai/ciris/api` 1252), xcframework (both slices, marker 0.5.199), desktop jar 1.5.199, `ciris-server` 0.5.196 range satisfied. Guard: no broad `except` may swallow a driving call, now that the client reports honestly.

## Android — three runs, three layers, all now ours or fixed
- **[33704781359](https://github.com/CIRISAI/CIRISAgent/actions/runs/33704781359)** — test server answered `/health` with `testMode=true` for the first time ever. Then died; the artifact had no logcat (`command -v adb` false on the ubuntu runner, fallback inside the guard) → fixed `b605d1869`.
- **[33706020778](https://github.com/CIRISAI/CIRISAgent/actions/runs/33706020778)** — logcat: `NoSuchMethodError: Composer.startReplaceGroup` at `CIRISApp.kt:355`. Our shell hand-pinned Compose 1.6.1; the client `.aar` targets 1.7.x through the CMP 1.7.1 accessors. **Ours.** Fixed `b36148526`: core Compose through the same accessors, resolves 1.7.5, guarded against hand pins.
- **[33708152999](https://github.com/CIRISAI/CIRISAgent/actions/runs/33708152999)** — no crash; **setup wizard ran on Android for the first time**, stopped on YOU: every input/click acknowledged, wizard didn't advance. `/input` acknowledges on *post* over a `MutableStateFlow` (latest-value-only). `35625ac44` reads every input back and screenshots every stage; the run in flight tests the hypothesis.

## iOS — attributed, ours, and a distribution problem
First app-side evidence (`debbe98fc` fixed the os_log predicate that matched `mobileassetd` instead of `iosApp`): the embedded Python dies on `import struct` — `_struct.framework` never embedded because `embed_native_frameworks.sh` read the **device** slice for a simulator build. Fixed `5f2bb8462` (stdlib half). The third-party natives exist only as iphoneos builds and `pydantic_core` has no cp310 simulator wheel anywhere → **CIRISAgent#1135** (plan), **CIRISServer#532** (iOS wheels).

Also fixed on the way: macOS teardown `kill -9`'d the simulator app for holding a *client* socket on :8080; stage timestamps were flush-time; a host backend on :8080 could satisfy the simulator's health check (now refused).

## Windows / Linux / macOS
Green throughout, apart from one intermittent agent-side chat timeout (`epistemic_humility_conscience` hanging; interact budget expires 15 s before the LLM timeout could rescue it) seen once on Windows and once on macOS — cross-platform, not client.

## Filed
- **CIRISClient#31** — vendor this gate against candidate builds; `/input` ack-on-post; `startIfEnabled` idempotency; AAR dependency metadata
- **CIRISServer#532** — iOS wheels
- **CIRISAgent#1135** — iOS simulator native extensions

Refs CIRISClient#28, CIRISClient#30

https://claude.ai/code/session_01JJy1BhjJKoRi6615eZHYzm


emooreatx:
Unblocked — **CIRISClient 0.5.198** is published and complete.

- `ciris-client-0.5.198.aar` (18,636,057 bytes) and `ciris-client-0.5.198.xcframework.zip` attached to the release
- 5 PyPI wheels including `py3-none-any`
- gate re-run against the **downloaded** AAR, not the build output: `ai/ciris/mobile: 5875`, `ai/ciris/api: 1252`, both modules present

`tools/fetch_client_artifacts.py --platform android` should resolve now.

## What it fixes

CIRISClient#30 — the `input_username` regression that blocked this adoption. 0.5.197 made `/input` refuse with 422 when no text sink is registered (the right rule: before it, `/input` answered `success: true` after a fixed delay whether or not anything received the text) and then registered no sinks anywhere. Six screens dispatched `textInputRequests` and all were working; the gate asserted a registration none performed. Every desktop text input went undrivable, not just `input_username` — you hit that one first because YOU is the first step.

Fixed structurally: `rememberInputSinks(...)` now sits beside each collector naming the tags it handles (21 across six files), and `check_ui_drivable.py` fails the build when a dispatched tag is undeclared. Mutation-verified.

## Worth waiting one more cut if you can

**0.5.199** is publishing behind it and changes what your gate can rely on, because while fixing the `/health` half of CIRISClient#28 I found the mobile automation surface was not the same surface as desktop:

- **`handleInput` returned `success: true` on Android and iOS regardless of whether a field was listening.** That is the same defect that produced #30, still live on the two platforms where automation is hardest — a green mobile setup step could have entered nothing. Now refuses identically.
- **`/undrivable` and `/state` were desktop-only.** They are now served from the shared handler on all five platforms. If you want the pre-flight I suggested on #30 — call `/undrivable` after a screen settles and fail with its contents rather than discovering gaps control-by-control — 0.5.199 is the first release where that works on mobile.
- **`/tree` carries `canClick`/`canInput` everywhere**, and `/element` stamps identically.

Also in 0.5.199, the `/health` unreliability you flagged as non-blocking. Two different bugs: Android logged "Server started" immediately after Ktor's `start(wait = false)`, which binds asynchronously, so a bind race read as an app fault (one `/health` answering, the next dying with `RemoteProtocolError` is that shape exactly) — readiness is now verified by connecting. iOS ran a blocking `accept()` on `Dispatchers.Default` and handed accepted connections back to that same small pool, so the loop could starve the handlers it spawned; `accept()` now owns a dedicated `Worker`.

Both are verified by construction and by CI compiling them, **not on a device**. Your nightly answering `/health` on Android and iOS is the actual proof, which is why CIRISClient#28 stays open until it does.

PR #1177, opened 2026-09-17
deps: ciris-client 0.5.219 -> 0.5.223 (consent is a question; drivers answer it) — see #1179

emooreatx:
Pins `ciris-client` 0.5.219 → **0.5.223** (CIRISClient `4130872`, tag `v0.5.223`). This PR pinned 0.5.220 first (gate 35181551033: Windows reset refused — **do not pin 0.5.220**), then 0.5.221 (gate 35199231784: Windows green, Linux and macOS lost the reset exit race — **do not pin 0.5.221 for desktop resets**), then 0.5.222 (gate 35220430920: **green on all five platforms**). History below.

## What 0.5.223 changes on top of 0.5.222 — and why this PR touches the drivers (#1179)

**Send traces is a question, not a switch.** Screen 2's replication consent was a switch defaulted ON; it is now Yes / No with no default, `btn_next` disabled until answered (the age band's shape), with the cost of declining — no capacity score, no commons credits — stated on the question. Be-scored and location appear once the answer is Yes. Unanswered reads as not-consented. **No wire change.**

**Driver contract change:** on screen 2 the tags are `trace_consent_yes` / `trace_consent_no`; `toggle_trace_opt_in` is gone from screen 2 (the YOU-step `AnnounceDecisionCard` keeps it). `dd88d65` updates `web_ui/__main__.py` and `mobile/test_cases.py` to wait for and click the answer per `trace_opt_in`; the reachability guard keeps its meaning. **This gate has not yet run against 0.5.223** — the driver change should get a look, then dispatch `Five-Platform Live QA` on this branch.

## What 0.5.222 changes on top of 0.5.221

**Desktop returns to Startup after a reset instead of exiting.** Desktop `restartApp()` is `exitProcess(0)`, and until 0.5.220 it never ran after a reset: the `java.util.prefs` throw (#56) skipped logout's completion callback, so every desktop reset left the app on Login — which is what this gate's `reset_took_effect` has asserted against all along (it reads "Login with `btn_local_login`, or Setup" one second after the confirm and does not relaunch a desktop; its "desktop respawns" comment is not true). Fixing #56 made the exit real; on 0.5.221 Linux and macOS lost the race, Windows won. 0.5.222 does what iOS already does: Startup → `retry()` → `startServer()` re-attaches to the still-serving backend (the launcher case, #61) or relaunches its own from the wiped home, so the next screen is a genuine first run and the #61 "node still running" line is visible on Login. Android keeps its AppRestarter relaunch. No wire change.

## What 0.5.221 changed on top of 0.5.220 (CIRISClient#61)

**Attached is the normal case.** 0.5.220 refused Reset when the backend was one the client did not launch, to keep a wipe from claiming success under a live node it could not stop (#55). Right harm, wrong condition: `desktop_launcher` starts the API and *then* spawns the JAR against it, so on every launcher-started desktop the client is `ATTACHED` — 0.5.220 refused Reset on the whole product, and this gate's reset leg is exactly that path (`.env` must lose `CIRIS_RUN_WITHOUT_AI` within 20 s; it never did).

0.5.221: Reset wipes regardless of ownership (the next backend boot is a first run, which is what the dialog promises and what `test_reset_device_flow` asserts), probes the attached node at reset time, and when it is still answering says so in the log and on Login — *"the CIRIS node at :4243 is still running with its previous state — restart it to finish"* — instead of refusing or reporting a clean success. Desktop still exits after the wipe as 0.5.219 did, because `--launch` binds :9091 directly and does not replace a running client. No wire change, no capability change.

## What 0.5.220 brought (still in this cut)

**The client can reach the brain it shows (CIRISClient#54).** `apiBaseUrl` honours `CIRIS_API_URL`, and `syncBackendFrom` moves a loopback default to `:8080` **iff a brain answers there**. An explicit `CIRIS_NODE_URL` still wins, so #1168's launcher behaviour is honoured and no longer load-bearing; the probe gate keeps this gate's bare-node desktops on `:4243`.

**One desktop auth cascade (CIRISClient#55–#59):** the `java.util.prefs` "Node has been removed" throw that skipped logout's completion callback (so the post-wipe restart never ran) is fixed (#56); a `410` on the OAuth hand-off is terminal and the node's `reason_id` is shown in-app with its recovery instruction (#57); an expired billing token renders "sign in again" with a Sign In control on Interact and Billing instead of 0 credits + Buy (#59) — pairs with the agent-side **401 `billing_auth_expired`** if that has landed; sidebar shows the generated version (#58). Also `/scroll` can name the nav rail (`container=nav_rail`); #42/#44 verified fixed on the YOU step.

## What was verified

- Release `v0.5.223` carries `ciris-client-0.5.223.aar` (18.8 MB) and `ciris-client-0.5.223.xcframework.zip` (114 MB) plus the localization bundle; PyPI wheel verified (`CIRIS-linux-x64-1.5.223.jar`). 0.5.222's release likewise carried `ciris-client-0.5.222.aar` (18.8 MB) and `ciris-client-0.5.222.xcframework.zip` (114 MB) — the two the "do NOT pin" notes in `requirements.txt` exist for — plus `ciris-client-0.5.222-localization.zip` (3.3 MB). Verified on PyPI: `ciris_client-0.5.222-py3-none-manylinux_2_17_x86_64.whl` (66 MB) downloads and carries `ciris_client/_artifacts/CIRIS-linux-x64-1.5.222.jar`.
- Client main CI green (all five jobs) on the tagged commit; `:shared:desktopTest` and the 126-test Python suite green locally on the fix.
- Client main CI green on the tagged commit; `:shared:desktopTest` and the 126-test Python suite green locally on the fix.
- Live evidence so far: 0.5.221's Windows leg passed the launcher-attached reset (`.env` gone); 0.5.222's change is the exit path only, so this run is the test for Linux/macOS. OAuth-410 and expired-token remain unit-level only.

## How to test it here

`Five-Platform Live QA` → `workflow_dispatch` on this branch with `client_preview_tag` left empty (uses the requirements pin).


emooreatx:
Five-Platform Live QA dispatched against this branch (requirements pin, defaults — all platforms, live key): https://github.com/CIRISAI/CIRISAgent/actions/runs/35181551033

This is the first end-to-end exercise of 0.5.220 outside the client's own gate — in particular the two paths that were only unit-verified there: the OAuth-410 refusal and the expired-billing-token "sign in again" state. I'll post the per-platform result here when it completes.

emooreatx:
**HOLD — do not merge. 0.5.220 has a desktop-product regression the gate just caught.**

The Windows leg failed `reset_took_effect`: `.env still records the install 20s after the reset was confirmed (factory reset stopped clearing it)`. From the client's own log in the artifact:

```
[PythonRuntime.desktop] A backend is already serving http://127.0.0.1:4243 — attaching to it
[CIRISApp] [WARN][onResetSetup] refusing: attached to a backend this app did not launch and cannot stop
```

That's CIRISClient#55's fix doing exactly what it says — and #55 was wrong about *where* that's acceptable. `desktop_launcher.py`'s docstring: *"3. The desktop app connects to the **running** CIRIS API server."* `ciris_engine/cli.py` starts the API, then spawns the JAR. So on **every launcher-started desktop the client is attached**, and 0.5.220's Reset refuses on all of them. The gate has the same topology, which is why it caught it; with 0.5.219 the gate was passing *through* the #55 bug (wipe under a live backend, then `desktop-setup --launch` restarted things).

Your comment 0b already anticipated this: *"Driving the real button is also how we find out if that reset ever stops clearing the flag."* It did.

**Fix belongs on the client, not the gate.** 0.5.221 should make the attached-and-live branch wipe local state as 0.5.219 did *and say the truth* — "erased; the node on :4243 still holds its old state until it restarts" — rather than refuse. That keeps #55's actual complaint (a reset that reported success and changed nothing) without killing Reset on the product. Tracking as a CIRISClient issue; this PR will move to 0.5.221 once that ships. Cancelling the remaining legs of run 35181551033 — the other two would fail identically.

Everything else in 0.5.220 (#54 brain routing, #56 logout throw, #57 OAuth 410, #59 sign-in-again, #58 version) stands and is not affected.

emooreatx:
**Gate 35199231784 on the 0.5.221 pin — interim, per leg:**

- **windows: green 8/8 + 3/3.** The reset leg that failed on 0.5.220 passes: `reset_device` → "the app restarted into first run" → `.env is gone`.
- **linux-android: linux reset FAILED; android reset passed (19 s).** Linux: `reset_device` OK, then 45 × "automation server not answering — the app is restarting; reconnecting", then `reset_took_effect` timed out. Android/iOS legs and the rest of the linux pass were fine.
- macos-ios: still running.

**Root cause (client side, not the pin, not this gate):** `reset_took_effect`'s desktop check is "Login with `btn_local_login`, or Setup", read ~1 s after the confirm — and the driver never relaunches a desktop (its comment "desktop respawns" is not true: the client's desktop `restartApp()` is `exitProcess(0)` with a "please restart manually" on stdout). It has passed on every nightly since 0.5.219 in **3–4 ms** because the desktop never actually exited after a reset: the `java.util.prefs` throw skipped logout's completion callback — the bug 0.5.220 fixed as CIRISClient#56. With #56 fixed the exit is real, and 1 s is a race: Windows won it this run, Linux lost it. Nothing about the wipe itself regressed — the `.env` assertion is the one that matters and it passed where it was reached.

**Fix, in 0.5.222:** desktop does what iOS already does after a wipe — return to Startup instead of exiting. `retry()` calls `startServer()`, which re-attaches to the still-running backend (the launcher case, #61) or relaunches its own from the wiped home, so the next screen is a genuine first run and the #61 "node still running" line is actually visible on Login. That is the better product behaviour independent of this gate, and it is the behaviour this gate has been asserting all along. Builds and unit tests are running now; I'll re-pin here when the release carries `.aar` + `.xcframework.zip`.

Suggested driver follow-up (not blocking): the "desktop respawns" comment in `reset_took_effect` should say what the desktop actually does, and a desktop that *does* exit should be a failure with its own signature rather than an `UNKNOWN`.

emooreatx:
Gate 35199231784 finished: **macos-ios failed with the identical signature** on the macOS reset leg (`reset_device` OK → 45 × "automation server not answering" → `reset_took_effect` timeout; the iOS half and everything before the reset were fine). So: two of three desktops lost the exit race, Windows won it. Same fix, 0.5.222, in flight.

emooreatx:
**Gate 35220430920 on the 0.5.222 pin: green on all five platforms** (https://github.com/CIRISAI/CIRISAgent/actions/runs/35220430920).

Per leg:
- **windows** — green (as on 0.5.221).
- **linux** — green. The reset leg that timed out on 0.5.221 now reads: `reset_device` (1.1 s) → "the app restarted into first run" → `.env is gone` → `reset_took_effect` **1.0 s**. Every Linux suite 7/7, 8/8, 3/3, 5/5.
- **android** — green; reset via AppRestarter relaunch, `reset_took_effect` 16 s.
- **macos** — green. Reset leg:
```
  the app restarted into first run
  /Users/runner/work/_temp/ciris-macos/.env is gone — the recorded choice went with it
 [OK] reset_took_effect (1069ms)
  the app restarted into first run
 [OK] reset_took_effect (27871ms)
```
- **ios** — green.

So the two things this PR's three pins were chasing are both now demonstrated live under the launcher topology: (1) Reset wipes on an attached backend instead of refusing (#61; 0.5.220's failure), and (2) the desktop stays up and lands on first run afterwards instead of exiting (0.5.221's Linux/macOS failure). The `.env` assertion — the one that matters for the with-AI pass — held on all three desktops.

Ready for review. The one driver-side note stands: `reset_took_effect`'s "desktop respawns" comment describes something the desktop client never did; now that it returns to Startup the assertion is honest, but the comment should say so.

emooreatx:
Pushed `dd88d65` + `de14b72`: pin **0.5.223** and the driver change it requires (screen 2's send-traces is now a required Yes/No — `trace_consent_yes` / `trace_consent_no`; `toggle_trace_opt_in` is gone from that screen). Details and the remaining steps for this repo are in #1179. The 0.5.222 result (green on all five platforms, run 35220430920) stands as the last verified state; the gate has **not** been re-dispatched on 0.5.223 from the client side — the driver edit deserves a look first, then `Five-Platform Live QA` on this branch with `client_preview_tag` empty.

Issue #1029, opened 2026-08-12
Adopt ciris-server 0.5.169 — the floor for #1028: login/health fixes you reported (never shipped until now), an unauthenticated privilege escalation closed, and the receive axis for delegated moderation


emooreatx:
**ciris-server 0.5.169 is cut and on PyPI** (9 wheels: 3 android, 2 macos, 2
manylinux, win_amd64, sdist). Substrate: **persist v30.11.0 / edge v15.22.0 /
verify v13.0.0**.

This is the floor for #1028 — and it carries three things that were finished but
**never shipped**, because 0.5.168 was written and never tagged. The last cut was
v0.5.167, so everything below reaches a released wheel for the first time here.

---

## 1. Two defects this team reported, both fixed — and they were the same defect

**A failed login logged nothing.** "This node has never heard of that user" and
"the stored hash did not match" were indistinguishable, and they have opposite
fixes. Every branch is now logged node-side (the wire responses are unchanged —
not leaking whether an account exists is correct), and the miss report carries
**how many certs were scanned**, which separates the two zeroes: `scanned=0`
means the node is reading a different database entirely; `scanned>0` means the
certs are there and none is named that.

The root cause, now pinned by test: `owner-hint` advertises
`first_name = name.split_whitespace().next()`, while `resolve_login` requires the
**whole** name exactly. A node whose owner is `jeff smith` advertises `jeff` —
and `jeff` is the one string login refuses. Matching stays exact on purpose:
accepting a first name makes login ambiguous the moment two people share one, on
the path where ambiguity means signing in as the wrong person.

**A folded agent reported as a bare NODE.** `/v1/system/health` is a substrate
path the node answers natively and never proxies, so on a folded port the client
hid the 22 cognitive services of the agent it was talking to — while the brain's
own port 404s the node surface. **Neither port served both meanings.** It is now
the union: node health is always the base, the brain's `cognitive_state` and
service map merge on top, and `agent.folded` / `agent.reachable` are separate
booleans because "no brain" and "brain did not answer" are different facts that
both used to render as a bare node.

Proxying the path wholesale would have answered the agent question and silently
lost node liveness.

## 2. Session tokens are verified at EVERY door

Worth calling out separately because it is exploitable, not cosmetic, and any
node binding `0.0.0.0` is exposed:

- `POST /v1/auth/refresh` accepted a **forged** session token and returned a
[...]

PR #1182, opened 2026-09-20
fix(config): home X is home X — one home's .env, never ~/ciris's

emooreatx:
## Home X is home X: the agent reads one home's `.env`, never `~/ciris`'s

A run launched with `CIRIS_HOME=X` and no `X/.env` booted on X's keys, identity and logs — and on **`~/ciris`'s database**. Two homes in one process.

### What happened (2.11.3, embedded server 0.5.204, from the logs)

- `main.py`'s boot loader kept its own list of `.env` candidates that always ended with `~/ciris/.env` and `/etc/ciris/.env`, and loaded *every* existing one (`override=False`). With no `X/.env`, `~/ciris/.env` supplied `CIRIS_DB_PATH`, `CIRIS_DATA_DIR`, `AUDIT_LOG_PATH`.
- `first_run.get_config_paths()` appended `~/ciris/.env` as a "legacy fallback" regardless of `CIRIS_HOME`, so first-run detection said *configured* for a home that had nothing in it.
- `env_utils.load_env_file()` defaulted to `./.env` — cwd, a third resolver.

Result: `ciris-persist Engine constructed (dsn=sqlite:////home/emoore/ciris/data/ciris_engine.db)` under `CIRIS_HOME=/tmp/…/agenthome`. The fresh actor + node keys met an already-claimed store: `root-user bootstrap evaluated outcome=AlreadyBootstrapped — no first-run claim PIN minted` and, in the same second, `no owner signer is on disk, so its owner-binding could not follow … Clause D is fail-closed`. Unclaimable and unowned at once — the three-key model (CC 3.4.7.3) doing exactly its job on an input it should never have been given. The wizard then ran (forced), could not claim, and `setup/complete` wrote ROOT WAs into the operator's real store.

### The rule

`get_ciris_home()` is the one resolver (managed `/app` → `CIRIS_HOME` → app sandbox → dev cwd → `~/ciris`), and **the `.env` this process reads is that home's `.env`**. `CIRIS_CONFIG_DIR` may put a config file ahead of it (the HA add-on keeps config apart from data). `/etc/ciris/.env` is a *system install's* file and is read only when the home is the implicit `~/ciris` default. Read path = write path (`get_default_config_path()` already resolved this way).

| | before | after |
|---|---|---|
| `CIRIS_HOME=X` | `X/.env`, `./ciris/.env`, `./.env`, `~/ciris/.env`, `/etc/ciris/.env` (merged) | `X/.env` |
| dev cwd (git repo) | `cwd/.env`, `~/ciris/.env`, `/etc/ciris/.env` | `cwd/.env` |
| installed (no home named) | `~/ciris/.env`, `/etc/ciris/.env` | unchanged |
| `CIRIS_CONFIG_DIR=C` | `C/.env` first, then the above | `C/.env`, then the home's |

### Changes

- `first_run.get_config_paths()` — the one list, per the rule above; `first_run.load_boot_env()` — loads it into `os.environ` (`override=False`, so an explicit `CIRIS_DB_PATH=… ciris-agent` still wins) and returns what it loaded.
- `main.py` — the 50-line private list is gone; it calls `load_boot_env()` and the `[CIRIS STARTUP] Config:` banner now names the file(s) actually loaded, so a log shows which config a run is on.
- `env_utils.load_env_file()` — default is the home's `.env`, not `./.env`.
- Tests: `test_explicit_home_is_the_only_home` (CIRIS_HOME=X with a configured `~/ciris/.env` → `[X/.env]`, and `is_first_run()` is True), `test_load_boot_env_reads_only_the_home` (`CIRIS_DB_PATH` from `~/ciris` never enters the process; `X/.env` does), `test_installed_mode_keeps_the_system_file`; the dev-mode test now asserts `[cwd/.env]`.

Behaviour change to note: a dev checkout no longer inherits `~/ciris/.env`. That was the same leak wearing a different hat (the `.git` cwd *is* the home in dev mode), and it is the mode a developer with an installed agent hits.

### Verified

- `tests/ciris_engine/logic/setup/test_first_run.py` 35 passed; `tests/test_main_helpers.py`, `test_node_only`, `test_resume_from_first_run`, `logic/config`, `test_path_resolution`, `test_logging_config`, `test_logging_backend_separation`, `routes/setup` — 372 passed.
- Live: `CIRIS_HOME=<empty dir> python main.py --adapter api --mock-llm` with a configured `~/ciris/.env` present → banner `Config: <home>/.env — NOT FOUND (first run will create it)`, `DB_INIT … target=<home>/data/ciris_engine.db`, zero mentions of `~/ciris` in the boot log.

### Seen in the same run, not fixed here

- `setup/complete` is not idempotent and does not dedupe by name: two POSTs 54 ms apart created two ROOT WAs named `qaadmin` → `auth.login.ambiguous_name` for every login after (#1029 territory).
- `setup/complete` ran on a node whose server had declared `setup/root` closed (`AlreadyBootstrapped`); the Python gate asks persist for a ROOT `wa_cert` and the server asks for an owner binding — two definitions of "already owned".
- `CIRIS_FORCE_FIRST_RUN` is re-evaluated on every status check, so setup stays "required" after completion.

https://claude.ai/code/session_01Xv1s8qrUoPoFX2MHZUPA15


emooreatx:
Reviewed, and pushed four fixes onto the branch (`43df5ddf0`) — the rule is right and the fix works. Verified the headline case directly: `CIRIS_HOME=X` (empty) with a configured `~/ciris/.env` present →

```
paths       : ['~/X/.env']
loaded      : []
CIRIS_DB_PATH leaked into env: False
is_first_run: True
```

## 1. The list changed its answer mid-boot

`get_config_paths()` gated the `/etc/ciris/.env` row on `not os.environ.get("CIRIS_HOME")` — and `ensure_ciris_home_env()` **exports `CIRIS_HOME`**, always, including when nobody named it. `main.py` calls `load_boot_env()` *before* that export and everything else asks *after*, so one process got two lists. Measured on the branch:

```
before ensure_ciris_home_env(): ['~/ciris/.env', '/etc/ciris/.env']
after  ensure_ciris_home_env(): ['~/ciris/.env']
```

For a system install whose only config is `/etc/ciris/.env` that is not cosmetic. Boot loads it and `CIRIS_CONFIGURED` enters the environment; `is_first_run()` then finds no config file in its shortened list, reaches its *"clear any stale CIRIS_CONFIGURED"* branch, **deletes the variable it just loaded**, and returns `True` — the wizard runs on a configured machine. Two answers in one process is this PR's own fault one level up.

The site-wide file belongs to the `~/ciris` home, so that is now the whole test: `ciris_home == Path.home() / "ciris"`. Same answer whenever it is asked.

## 2. Both failing shards were real

- **`test_is_first_run_with_user_env`** — a `CIRIS_HOME` leaked by `test_resource_monitor.py` decided the answer. The leak is the actual bug: two of its five restore sites do `if original_env: os.environ[...] = ...` with no `else`, so a temp dir that no longer exists outlives the test. Restoring an unset var means deleting it. The old unconditional `~/ciris` row is exactly what hid this — the PR didn't break the test, it exposed it.
- **`test_loads_env_priority_order`** — patched `main.Path` / `main.load_dotenv` and reloaded `main`. Those attributes are gone by design. Rewritten against the seam the guarantee lives in now: `load_boot_env()` loads the home's `.env`, in order, `override=False`, and an explicit `CIRIS_DB_PATH` in the real environment still wins.

## 3. The failure banner named two files nothing reads

`main.py`'s *"Or mount a .env file at:"* printed `~/.ciris/.env` and `./.env`. The first is the keys/secrets directory this module's own docstring says is never config; the second stopped being a home in this PR. An operator who followed either got a file the agent ignores. It now prints `get_config_paths()`.

## 4. Two regression tests

`TestOneAnswerThroughoutTheBoot`: the list is identical before and after the `CIRIS_HOME` export, and a named home elsewhere never reads the system file. Both verified to go red with the old condition restored.

## Verification

506 passed across `logic/setup`, `logic/config`, main CLI/helpers, `test_resource_monitor`, and every suite that asserts on config resolution; `first_run.py` and `env_utils.py` mypy-clean (the two remaining mypy errors are the pre-existing wallet-provider import skips).

## Out of scope, worth an issue

Three resolvers still disagree with the one-home rule: `adapter_manager.py:1446` falls back to `~/.ciris/.env` (a fourth location, and the keys dir again), `node_only.py:107` keeps its own `env_file_path()`, and `resume_helpers` / `setup/complete` / `billing` / `resource_monitor` each `load_dotenv(..., override=True)` against a path they resolve themselves. None of them is the bug this PR fixes — they all derive from `CIRIS_HOME` or a passed path — but "the `.env` this process reads is the home's `.env`" isn't structurally enforced until they route through `get_config_paths()` too.

